PR #1357, opened 2015-01-10
Libports

ehmry:
I'm not using these libraries yet but here are the initial definitions if anyone else wants to use them.


nfeske:
Thank you for the pull request! I would separate the change of ```tool/ports/mk/install.mk``` to a separate commit though.

ehmry:
split it

nfeske:
Thanks! I just merged the patches into staging and will close the issue once they arrive at the master branch.

PR #1389, opened 2015-02-01
sqlite: native SQLite OS layer

ehmry:
This patch makes SQLite run with less Unix cruft and without any fcntl calls. File system access uses the Libc VFS. It has a strict dependency on Timer and RTC sessions, I know RTC isn't supported yet on ARM but I'm worried it could cause problems later if it can't timestamp data. 
EDIT: Rtc is optional now.

At some point it will need file locking support and when the VFS detaches from Libc I think there are some optimizations that can be made.

The test is pretty heavy, I don't know offhand if its IO bound or not, but it might be usefull for stress testing other components.


ehmry:
Is there a convient way to download a single source file without defining a port?

chelmuth:
You may add the single file to your sqlite port. For illustration, I added a single file to zlib in the following
```diff
diff --git a/repos/libports/ports/zlib.port b/repos/libports/ports/zlib.port
index 60367db..20d1916 100644
--- a/repos/libports/ports/zlib.port
+++ b/repos/libports/ports/zlib.port
@@ -1,10 +1,13 @@
 LICENSE   := zlib
 VERSION   := 1.2.8
-DOWNLOADS := zlib.archive
+DOWNLOADS := zlib.archive robots.file
 
 URL(zlib) := http://downloads.sourceforge.net/project/libpng/zlib/$(VERSION)/zlib-$(VERSION).tar.gz
 SHA(zlib) := a4d316c404ff54ca545ea71a27af7dbc29817088
 DIR(zlib) := src/lib/zlib
 
+URL(robots) := http://genode.org/robots.txt
+SHA(robots) := c47ccf1a49c24cc5842430aa75c72ef491292412
+
 DIRS := include/zlib
 DIR_CONTENT(include/zlib) := $(addprefix src/lib/zlib/,zconf.h zlib.h)
```

ehmry:
Thanks, that cut the SLOC in half.

ehmry:
I'm going to close this and pull again when I feel there aren't any more libc issues.

chelmuth:
@ehmry I suggest you open a regular issue for your development as we do not use pull requests often. The issue can track all discussions and development steps and stays open until the solution is merged into our master branch. IMO opening and closing pull requests frequently increases the noise. We follow the convention to tag our commits related to an issue with `Issue #XXX` and conclude with a final commit stating `Fixes #XXX`. What do you think?

ehmry:
I think you're right, and that way should be cleaner for myself as well.

PR #1391, opened 2015-02-02
libc-vfs: support access

ehmry:


ehmry:
Right now I'm pulling the access mode from the least signicant bits, ---|---|rwx, should I look at the 'user' bits or the 'other' bits? Or is this why access wasn't implemented?

cnuke:
Actually, we do not care about the permission bits at that point because the permission to access the file system is enforced by the routing policy. Therefore, access merely should show if a path/file is accessible and should ignore any bits. It stands to reason if we really need to support UNIX permissions and up to now this question was answered with *no* since there is no notion of users on Genode.

chelmuth:
Ah okay, so it boils down if any third-party application uses `access()` to implement program logic, right?

cnuke:
Yes and if we want to support that we would probably have to do more than only checking the permissions in `access()`.

ehmry:
Yea, I realize now that checking bits is pretty pointless.

PR #1433, opened 2015-03-05
signal: fix infinite loop

ValiValpas:
There is an infinite loop in signal.cc. The comment mentions that the loop shall be exited after a second try, which is controlled by setting the try_again variable to true after the first try. However, the variable is reset to false on each iteration.


m-stein:
I find the semantics of the ``try_again`` flag pretty irritating. I expect it to express wether something should be tried again and not the other way round. Furthermore, while working on an alternative, I've found a false positive  in the signal test that comes into play when fixing the infinite loop.

[This branch](https://github.com/m-stein/genode/tree/1433_signal_fix_infinite_loop) fixes the above problems and this issue.

Making the ``retry`` tool in ``platform_env.h`` public and using it for such problems in general would be an even better approach I think. Any opinions on that?

PR #1445, opened 2015-03-12
hw: Let CPU implementation specify mtc region size

reet-:
For the base-hw port to x86_64, we need the mode transition control region to span two pages. Currently, the mtc size is hard-coded to the minimal mapping size of the page tables implementation (which is 4096 bytes on x86).

Decouple the mtc size from the minimal page tables mapping size and let the concrete CPU implementation specify it instead.


nfeske:
Merged to staging.

PR #1448, opened 2015-03-13
Port base-hw kernel to the Intel x86_64 architecture

reet-:
This changeset ports the base-hw microkernel to the Intel x86_64 architecture. The port passes the same base/run tests as the hw_pbxa9 target.

The starting fixup commits are intended to be squashed with the commit referenced in the respective git subject line. Commit 5fdb6ed can be dropped as soon as the corresponding fix hits the staging branch.

Current limitations:
- x86_32 is not supported
- No SMP support


nfeske:
Very impressive line of work! :+1:

m-stein:
Cool stuff. Currently wading through the commits...

skalk:
Once again: congratulation for the clean and considerable work!

I have taken the liberty to rebase your work again regarding the current staging branch that includes the solution for the "core bootstrap / exception handling initialization" problem. This solution differs significantly from the quickfix that I have sent to you beforehand, so more fixups were necessary. Moreover, I have added a minor fixup regarding coding-style issues (mainly: spaces/tabs mixture and 80 column violations). You can see the result on this branch: https://github.com/skalk/genode/tree/hw_x86_64

Before merging this to our staging branch we would like to squash the commits a bit. Normally we do not do commits with such a fine-granularity. Are you ok with this, or would you prefer to do this on your own?

skalk:
One open issue that I have forgotten to mention is: currently, there is no user-land timer for hw_x86_64 available. Therefore most scenarios (beyond base/run) fail to start on this platform. Normally we integrate every new platform into our nightly test-routines. Most other x86 drivers will work out-of-the-box for hw_x86_64, but for the user-land timer driver we only have a PIT driver for x86 available. As long as the kernel uses this, we have a problem here. Due to the fact that the kernel has to use another timer when e.g. supporting more than one CPU, we think it would be best to change the kernel-timer to use the local APIC timer. Do you agree?

m-stein:
I think specializing the kernel configuration because of the PIT limitation should be treated as quickfix only. A better way to solve this would be to apply such a limitation when scheduling in 'Cpu::exception'. If you agree with me I'd create an appropriate fixup (applicable also to the APIC suggestion).

Kensan:
First of all thank you for all the kind words and the rapid feedback. We are happy that the code is to your liking ;)

We had a separate commit regarding the "core init" upstream fix (cb888f1...) but did not include it since it was not yet available in the staging branch. Thanks for doing the necessary fixups!

With regards to the detailed commit history: our line of reasoning is that the current history provides additional context and the motivation for the concrete changes. Obviously the fixup commits should be combined with the corresponding changes. What other commits would you like to squash together?

We chose to implement the kernel timer using the PIT since for the APIC timer one has to measure the frequency to use it. Using the APIC timer as kernel-timer is certainly viable and we can do the implementation if you wish..

Ideally we would like to see a userland timer along the lines of the NOVA or Fiasco platform timer. Our final goal is to port base-hw to the [Muen Separation Kernel](http://muen.sk/). In this scenario there are potentially multiple base-hw subjects running, which means we are not able to spare a hardware timer device for each instance.

skalk:
@Kensan:

Regarding the detailed commit history: beside squashing the fixup commits, the idea was to group for instance commits for timer support, for PIC support, or to squash the different commits for 1-6 arguments syscalls into one.

With regard to the timer implementation: propably it will be good to have a generic solution for all base-hw platforms in the line of NOVA and Fiasco.OC. Nevertheless, we need some time for thinking about this API change. Therefore, from our perspective it would be best to use the local-APIC for the time-being, and re-use the available PIT timer driver for the user-land. If it is within the bounds of possibility for you to do the implementation, I would be delighted. Otherwise, we need to time the development for the upcoming week, but I could imagine you are faster in it ;-).

nfeske:
Let me explain our usual practice with respect to the commit history.

While working on a topic branch, we use to commit many small intermediate steps. This is useful to keep track of the line of thoughts during development. This history is regarded as volatile. That is, it is not set in stone. So we do not need to spend too much thoughts when committing something.

Once the work on the topic is completed and we are going to integrate the topic branch into the mainline development, we curate the topic-branch history so that a short and well-arranged sequence of commits remains. In many cases, the entire topic branch is squashed into a single commit. The goal behind this curating step is to make the mainline history document the progress at a level of detail that is meaningful for the users of the framework. For the mainline history, we want the achieve the following:

* The relationship of a commit with an issue at our issue tracker should be easily possible. For this reason, we add GitHub's annotations ```issue #n``` and ``` fixed #n``` to the commit messages.
* Revisiting the history between Genode releases should clearly reveal the changes that potentially interest the users. I.e., when writing the quarterly release notes, we go through the history and base the document on the information contained in the commit messages. This works best if each topic is comprised by a few commits with meaningful descriptions. This becomes hard if the history contains too many details. Think of a signal-to-noise ratio.
* Each commit should represent a kind of "transaction" that can be reviewed independently without knowing too much context. This is hardly possible if intermediate steps that subsequently touch the same code are present as individual commits.
* It should be easy to selectively revert individual topics/features using ```git revert``` (e.g., when trouble-shooting). This is simple when each topic is represented by one or just a few commits.

Given this rationale, your topic branch presents quite many opportunities for simplification. For example, instead of maintaining a sequence of patches "implement Pic::... function", it would be preferable to have a single commit "Pic implementation". Rather than documenting the sequence of development in detail, it would document what is there. Whereas the former may be interesting by the original developer, the latter has more value to the consumers of the software.

That said, I do not want to force this practice on you. If you insist to retain your detailed history, we will keep it as is.

Kensan:
Thank you for the in-depth explanation of your workflow and the motivation behind it. Our main concern was, that squashing the entire x86_64 port into one or two commits would leave us with a history that is too coarse for understanding particular implementation choices and details. Since this is not the intention, we are certainly fine with you simplifying the commit history as you see fit in order for it to conforms to your practice and standards.

As for the APIC timer: we will try to come up with a suitable implementation in the coming days.

Kensan:
This updated pull request addresses the discussed issues and extends the base-hw/x86_64 implementation in order to support scenarios that require a userlevel timer (e.g. run/signal). Additionally the changes have been rebased on the current staging branch (48d3980...).

As requested, we have redone the kernel timer implementation to use the APIC timer instead of the PIT. In the course of doing so we have also upgraded the PIC implementation to use the local and I/O APIC for interrupt management. This should simplify the addition of SMP support in the future, since the APIC is a requirement for the implementation of x86 multicore support.

With some additional changes, that are also included in the update, it is now possible to run the demo scenario on base-hw/x86_64 in QEMU (and Bochs).

chelmuth:
:+1:

nfeske:
I am amazed about your work. It is so cool that you can run full scenarios including user-level device drivers on the hw kernel on x86 now. I am all for merging the current patch series into staging (with the slight change of the timer.mk location). @m-stein or @skalk would you confirm?

m-stein:
Its cool to see this code. I need a little time to go through all commits.

skalk:
@nfeske : I've again rebased it to staging, made some minor coding style corrections within the new local APIC implementation, squashed a commit, and corrected the problem with the user-land timer. You can find the result within my hw_x86_64 branch.

m-stein:
@Kensan, @nfeske:

To me, it was a pleasure to read the code. I've added 2 minor fixes. One merely regarding style and one that makes the specialization of ``Irq_session_component.cc`` unnecessary. Are these fixes ok for you?

m-stein:
PS: [Branch](https://github.com/m-stein/genode/tree/1448_hw_x86_64)

m-stein:
@nfeske:

Rebased my branch 1448_hw_x86_64 onto @skalk's branch

Kensan:
Thanks for the prompt feedback and the fixups! @reet- and I reviewed your changes and from our side they look fine.

nfeske:
I will merge the branch to staging as soon as the current batch of pending commits on staging arrive in master.

reet-:
The current code has an issue with edge-triggered interrupts because it is generally problematic to mask them. To quote the Intel Intel 82093AA I/O Advanced Programmable Interrupt Controller (IOAPIC) specification, section 3.4.2 on Redirection Table Entries:

*When this bit (editors note: the "Interrupt Mask" bit) is 1, the interrupt signal is masked. Edge-sensitive interrupts signaled on a masked interrupt pin are ignored (i.e., not delivered or held pending)*

There is a race between the userland timer triggered unmask and the programmed one-shot interrupt. If the timer fires while the line is still masked, the interrupt is lost and the userland code sleeps forever waiting for the next IRQ to happen. While debugging the userland timer code, we have seen timeouts of 0.25 ms which seems rather small.

We therefore propose the fix in commit 9e9953f of our [hw_x86-edge-triggered](https://github.com/codelabs-ch/genode/tree/hw_x86-edge-triggered) branch. Since PCI interrupts are always level-triggered and edge-triggered is (mainly) used for legacy ISA IRQs, it should be a viable solution.

m-stein:
@reet-, @nfeske  
That's a pretty cool fix. Admittedly, I didn't realize that the signal-based IRQ approach in hw would manage edge-triggered IRQs out of the box as I didn't depend on such IRQs by now :)

I re-based your commit to my 1448_hw_x86_64 branch that contains Stefan's and my review commits and added two minor style fixes. Does this look good to you?

reet-:
@m-stein 

I'm glad you like our fix for the edge-triggered interrupt problem!

The changes look good expect for the following minor nitpick: There is a typo in commit 6622b01: wether vs. whether (both in code and commit message).

m-stein:
@reet- 
Oh, thank you for the hint :) Fixed it.

reet-:
On 03/23/2015 01:17 PM, Martin Stein wrote:
> @reet- 
> Oh, thank you for the hint :) Fixed it.

great thanks!

chelmuth:
Amazing, hw_x86_64 went pretty smooth to our nightly build-and-test cycle. Qemu/hw_x86_64 succeeded for 24 of 28 tests. The remaining issue is handling of MMIO regions as could be seen in the log of `run/libc_ffat`
```
[2015-03-26 23:50:46] [init -> ahci] Found AHCI HBA (Vendor ID: 8086 Device ID: 2922 Class: 00010601)
[2015-03-26 23:50:46] I/O memory [febf1000,febf2000) not available
[2015-03-26 23:50:46] Local MMIO mapping failed!
[2015-03-26 23:50:46] [init -> ahci] C++ runtime: Genode::Parent::Service_denied
[2015-03-26 23:50:46] [init -> ahci] void* abort(): abort called
```
Nevertheless, I'm going to cleanup and merge staging to master today.

Kensan:
Thanks for the feedback. It is great to hear that the majority of tests pass.

As for the MMIO issue: currently only the MMIO regions specified in the x86-specific [platform_support.cc](https://github.com/genodelabs/genode/blob/staging/repos/base-hw/src/core/spec/x86/platform_support.cc) file are given to the MMIO allocator. Adding the missing ranges to the _mmio_regions should do the trick for the short term.

A long term solution could be to adjust the MMIO region allocation for x86 to either determine the range dynamically or implement the approach taken by [base-nova](https://github.com/genodelabs/genode/blob/master/repos/base-nova/src/core/platform.cc#L342).

PR #1569, opened 2015-06-08
base-hw: add Zynq-7000 support

ValiValpas:
Supporting Qemu (hw_zynq), parallella board (hw_parallella) and zedboard
(hw_zedboard).

Device drivers: nic, gpio, vdma, i2c

Thanks to Timo Wischer and Mark Albers for their contributions.

### DIFF
--- a/repos/base-hw/lib/mk/platform_parallella/core.mk
+++ b/repos/base-hw/lib/mk/platform_parallella/core.mk
@@ -1,0 +1,16 @@
+#
+# \brief  Build config for Genodes core process
+# \author Johannes Schlatow
+# \date   2014-12-15
+#
+
+# add include paths
+INC_DIR += $(REP_DIR)/src/core/include/spec/parallella
+INC_DIR += $(REP_DIR)/src/core/include/spec/xilinx
+
+# add C++ sources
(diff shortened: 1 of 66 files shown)


ValiValpas:
@skalk:

First of all, thanks for your appreciation.

Yes, I know about the GCA and Timo and Mark are happy to sign it.

We will certainly be working with the zedboard and the parallella for a while, thus we could play such a role for sure.

About splitting the commit: I don't have any objections. If I understand you correctly, you'll be taking care of this, or do you need me to take action here?

skalk:
@ValiValpas: that is good news regarding the GCA, and especially your committment to maintain the board support for now.

Regarding the split of the commit: to be honest I would like you to do this, because otherwise it might take a while until I'm able to do so (time constraints), I can only test the QEMU variant, and lastly because, the majority of commits would loose their actual authorship. So if you do not mind I would be happy if you would continue to care for the commits primarily.

ValiValpas:
@skalk: Alright then. It will probably take me a couple of days as well. Stay tuned!

ValiValpas:
Closing this pull request because it is superseded by #1577

PR #1587, opened 2015-06-16
Add nullptr error handling in part_blk, fix #1586

### LINKED ISSUE #1586: Missing nullptr check in part_blk
Patch pending.

taruti:
This fixes an unchecked potential null pointer reference.


chelmuth:
The fix is in master.

PR #1577, opened 2015-06-12
Add Zynq-7000 support for QEMU

ValiValpas:
This is a follow-up of #1569. Now only containing the basic support of zynq-based platforms provided by qemu.

I ran the autopilot tests with the following results:
```
--- platform hw_zynq ---
hw_zynq:             util_mmio              -> OK (0:34)
hw_zynq:             ldso                   -> OK (0:37)
hw_zynq:             timer                  -> OK (0:20)
hw_zynq:             lwip                   -> OK (0:11)
hw_zynq:             rm_fault               -> OK (0:07)
hw_zynq:             rom_blk                -> OK (0:06)
hw_zynq:             tar_rom                -> OK (0:06)
hw_zynq:             noux                   -> OK (1:18)
hw_zynq:             noux_net_netcat        -> ERROR (0:29)
hw_zynq:             libc_ffat              -> OK (0:00)
hw_zynq:             libc_vfs               -> OK (0:07)
hw_zynq:             timed_semaphore        -> OK (0:05)
hw_zynq:             signal                 -> ERROR (3:23)
hw_zynq:             sub_rm                 -> OK (0:03)
hw_zynq:             python                 -> OK (0:00)
hw_zynq:             l4linux                -> UNAVAILABLE
hw_zynq:             lx_hybrid_ctors        -> UNAVAILABLE
hw_zynq:             lx_hybrid_exception    -> UNAVAILABLE
hw_zynq:             lx_hybrid_pthread_ipc  -> UNAVAILABLE
hw_zynq:             moon                   -> OK (0:12)
hw_zynq:             thread_join            -> OK (0:05)
hw_zynq:             failsafe               -> OK (0:10)
hw_zynq:             netperf_lwip           -> OK (0:00)
hw_zynq:             netperf_lwip_usb30     -> OK (0:00)
hw_zynq:             netperf_lwip_bridge    -> OK (0:00)
hw_zynq:             netperf_lwip_wifi      -> OK (0:00)
hw_zynq:             netperf_lxip           -> OK (0:00)
hw_zynq:             netperf_lxip_usb30     -> OK (0:00)
hw_zynq:             netperf_lxip_bridge    -> OK (0:00)
hw_zynq:             netperf_lxip_wifi      -> OK (0:00)
hw_zynq:             l4linux_netperf        -> UNAVAILABLE
hw_zynq:             l4linux_netperf_usb30  -> UNAVAILABLE
hw_zynq:             l4linux_netperf_bridge -> UNAVAILABLE
[...]

PR #1599, opened 2015-06-22
Add Zynq-7000 support for QEMU (refactored)

ValiValpas:
This is a follow-up of #1569 and #1577 addressing the comments of @m-stein and @skalk.

In particular, I refactored the part that still made heavy use of preprocessor defines.


m-stein:
@ValiValpas 
Thank you for taking the time to apply our comments :-) After a quick look I've merely noticed some minor style issues. I'll push some fix-ups tomorrow.

m-stein:
@ValiValpas 
My branch [1599_Add_Zynq_7000_support_for_QEMU](https://github.com/m-stein/genode/tree/1599_Add_Zynq_7000_support_for_QEMU) re-bases your commits to our master branch and adds several fix-ups. I hope you're fine with my modifications?. I've marked your NIC refactoring as a fix-up too as it should be squashed into the original commit when finally merging. This way you're not the git author anymore but you nevertheless get mentioned in the commit message. You should anyway be mentioned as co-author in the documentation header of every file you adapted.

After all, there are still some issues left:
* I can't compile any network tests as you have missed to take over some files (marvel_phy.cc as example) from the original pull request. As these files contain lots of preprocessor defines and thus also need a more time-intensive clean-up, I would be glad if you could fix this.
* I haven't reviewed the NIC commit yet because of the above issue
* The user-land timer still shows bad behavior on ``run/signal`` at least

When fixing these things, please don't open up a new follow-up but simply put a ``Ref #1599`` in your commit messages (causes commits to be shown in this thread). This way it is much easier for others to follow the discussion and the development in the background. Please also use my [1599_Add_Zynq_7000_support_for_QEMU](https://github.com/m-stein/genode/tree/1599_Add_Zynq_7000_support_for_QEMU) branch as base for your new commitments to simplify synchronization of our collaboration.

ValiValpas:
I assume the problem with run/signal is that the triple timer counter (TTC) I used for implementing the platform timer is only a 16bit timer. Thus the maximum possible timeout is approx. 300µs (with maximum prescaling).
Maybe we should use the Cortex A9 Global Timer instead.

m-stein:
@ValiValpas The problem with the global timer is that resides on the same physical page as the private timers. The private timers, however, are used by the kernel for scheduling and must therefore be mapped in core only. Also using another timer in the kernel is no good idea in my opinion. We would have to find an alternative SMP-aware timer on every Cortex-A9 platform.

m-stein:
Maybe you can use the 24 bit watchdog timer or cascade the 2 triple timers instead?

ValiValpas:
Well, after a good night's sleep I noticed that I was totally wrong about my previous post. In other words, there was a PEBKAC issue with my prescaling implementation. Fixed it and `run/signal` is working properly now.

m-stein:
Cool that you've found the problem with the timer. The fix-up looks good. Anyhow, if I didn't miss a thing, the maximum timeout is now ~31 ms which is still pretty low. We should keep this in mind if there are future performance/timing issues on Zynq.

nfeske:
@m-stein I think that 31 milliseconds are just fine. Please keep in mind that the timer driver is shared my many components. In most scenarios, there is at least one component (e.g., nitpicker) that triggers timeouts at a rate of 10 milliseconds. Only when the system is idle, the maximum timeout comes into effect. And even in this case, the costs for handling a timeout every 31 milliseconds is negligible, isn't it?

m-stein:
Likely you're right... I'm a little bit overcautious :)

m-stein:
@ValiValpas 
I fetched your current state, merged the fix-ups to make re-basing easier, re-based the commits then and added some minor style fixes. As we shouldn't wait until re-basing gets more difficult and there are still open issues with the NIC commit, I would strongly vote for merging only base- and timer-support (https://github.com/m-stein/genode/commit/58b7b11296e5b5fa94802d23b1631b644a129bb7, https://github.com/m-stein/genode/commit/7f56e545947a8810537837d79121c0347fb65afb, https://github.com/m-stein/genode/commit/a1cce903565b6e4346538687d51f5591ba578ee0) into mainline and creating a new issue for NIC support. Do you agree with this?

Please use my branch [1599_Add_Zynq_7000_support_for_QEMU](https://github.com/m-stein/genode/tree/1599_Add_Zynq_7000_support_for_QEMU) for amendments.

ValiValpas:
@m-stein 
Agreed.

m-stein:
@chelmuth Could you please merge https://github.com/m-stein/genode/commit/58b7b11296e5b5fa94802d23b1631b644a129bb7, https://github.com/m-stein/genode/commit/7f56e545947a8810537837d79121c0347fb65afb, https://github.com/m-stein/genode/commit/a1cce903565b6e4346538687d51f5591ba578ee0 ?

chelmuth:
Merged three commits into staging. I'll also enable the platform in the buildbot.

m-stein:
@chelmuth Thanks!

m-stein:
@ValiValpas Sorry for messing up the author in the Zynq-related commits. In mainline, this will be corrected.

m-stein:
@chelmuth Added two fix-ups https://github.com/m-stein/genode/commit/0f04e329d70a0f6b91dce1f5203d2e2d2989cb38 and https://github.com/m-stein/genode/commit/cf5fa5bf465636ac28cf9c5fbeb7eb7304328077.

ValiValpas:
Great. Let me know if I can be of any help with the NIC support.

m-stein:
@ValiValpas I moved the NIC discussion to the new issue #1667.

PR #1708, opened 2015-09-21
Run: Support loading images via iPXE

Kensan:
These changes add a run script to support deploying images using [iPXE](http://ipxe.org/). iPXE is an open source network boot firmware that, among other things, supports booting ISO images via HTTP directly from a web server.

The following build.conf/RUN_OPT parameters can be used to configure the iPXE/HTTP setup:

--load-ipxe-base-dir: This parameter specifies the base directory of the HTTP server from which the target machine downloads the files, e.g. /srv/www.
--load-ipxe-boot-dir: The directory relative to iPXE base dir which contains the iPXE chainload configuration and all necessary files.

The target machine is expected to request the following iPXE configuration via HTTP: http://${HOST_URL}/${ipxe-boot-dir}/boot.cfg

This can be achieved by building iPXE with the following embedded script:

    #!ipxe
    dhcp
    chain http://${HOST_URL}/${ipxe-boot-dir}/boot.cfg

For additional information on iPXE an embedded scripts please refer to the [iPXE website](http://ipxe.org/howto/chainloading#breaking_the_loop_with_an_embedded_script).

NOTE: the webserver serving the ISO image must support ranged requests (see [here](http://forum.ipxe.org/showthread.php?tid=7295&pid=10482#pid10482)).

### DIFF
--- a/tool/run/boot_dir/fiasco
+++ b/tool/run/boot_dir/fiasco
@@ -144,4 +144,10 @@ proc run_boot_dir {binaries} {
 
 		generate_tftp_config
 	}
+
+	if {[have_include "load/ipxe"]} {
+		create_ipxe_iso_config
+		update_ipxe_boot_dir
+		create_symlink_for_iso
+	}
 }
--- a/tool/run/boot_dir/foc
+++ b/tool/run/boot_dir/foc
@@ -183,6 +183,12 @@ proc run_boot_dir_x86 {binaries} {
 
 		generate_tftp_config
 	}
+
+	if {[have_include "load/ipxe"]} {
+		create_ipxe_iso_config
+		update_ipxe_boot_dir
+		create_symlink_for_iso
+	}
 }
 
 
--- a/tool/run/boot_dir/nova
+++ b/tool/run/boot_dir/nova
@@ -93,4 +93,10 @@ proc run_boot_dir {binaries} {
 
 		generate_tftp_config
 	}
+
+	if {[have_include "load/ipxe"]} {
+		create_ipxe_iso_config
+		update_ipxe_boot_dir
+		create_symlink_for_iso
+	}
 }
--- a/tool/run/load/ipxe
+++ b/tool/run/load/ipxe
@@ -46,3 +46,23 @@ proc create_symlink_for_iso { } {
 proc update_ipxe_boot_dir { } {
 	exec ln -sfn [pwd]/[run_dir] [load_ipxe_base_dir]/[load_ipxe_boot_dir]
 }
+
+##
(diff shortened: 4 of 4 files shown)


chelmuth:
:+1: I like the nifty trick to sanboot the ISO. In the past, I had several issues with multiboot and iPXE/gPXE.

chelmuth:
I tested the patches and after some tinkering it generally works but sometimes hangs. Also on NOVA, several IOMMU faults are logged as iPXE seems still to be running in the background. So, we still have no silver bullet for network booting ;-)

Additionally, I think it's worth to mention that both `load/ipxe` and `image/iso` are required for NOVA, Fiasco, and Fiasco.OC like follows
```
RUN_OPT  = --include boot_dir/foc
RUN_OPT += --include image/iso
RUN_OPT += --include load/ipxe
RUN_OPT +=   --load-ipxe-base-dir /var/lib/tftpboot
RUN_OPT +=   --load-ipxe-boot-dir /x86
```
This configuration is compatible with my TFTP config.

BTW, would you mind if I squashed the _Add support_ commits into one?

chelmuth:
I almost forgot to ask: Do you have an approach to boot hw similarly?

Kensan:
What hangs are you observing? I tested with autopilot on NOVA and all the scenarios seemed to work ok on my hardware (Lenovo T430s).

Missing image/iso should prompt a failure message by the **create_ipxe_iso_config** function in the ipxe run script. I agree that I should have mentioned that fact more prominently somewhere.

> BTW, would you mind if I squashed the Add support commits into one?

No, I am fine with it. In that case maybe it would make sense to rework a047abd so all iPXE run script  changes are in one commit and the second commit simply adds support to the different platforms. I can do the cleanup and resubmit if you agree.

> Do you have an approach to boot hw similarly?

Yes, the motivation behind adding iPXE network boot to the run tool was autopilot support for hw_x86_64_muen. I intend to submit these changes next, see branch [hw_x86_64_muen-autopilot](https://github.com/codelabs-ch/genode/commits/hw_x86_64_muen-autopilot). Note that booting base-hw via iPXE is simpler since we do not need to assemble the ISO but can directly boot bender and load the kernel as a module (see f3cea05).

chelmuth:
> What hangs are you observing? I tested with autopilot on NOVA and all the scenarios seemed to work ok on my hardware (Lenovo T430s).

I was testing on a two year old notebook and in 30% of the boots the test execution just stopped in the middle of business. This was not tied to NOVA but also happened on Fiasco.OC. I'll merge the commit nevertheless as this may be an issue with my hardware only.

PR #1825, opened 2015-12-17
I.mx53 tablet demo

zhouyiteng:


nfeske:
@zhouyiteng What is the background of this pull request? It seems to be spuriously opened. May be close it?

Issue #987, opened 2013-11-29
lighttpd: stucks when file size exceeds ~260 KB


skalk:

gwitmond:
The cause is a limitation to write at most MAX_WRITE_LIMIT bytes at a time in: lighttpd/src/connections.c:627

    network_write_chunkqueue(srv, con, con->write_queue, MAX_WRITE_LIMIT);

However, the return code '1' suggest more writing, however that doen't happen.

It explains the apparent 255.57K limit. It's the size of the body without the http-headers. Add them up and it should be a perfect 256*1024.

The quick fix is to change that limit of 256K to almost infinite: SSIZE_MAX or even SIZE_MAX. I've tested that it works for static sites. No guarantees for http-chunk encoded output, however.

skalk:
@gwitmond cool, thank you for discovering the root cause!

ssumpf:
@gwitmond: :+1:  This bug has been around for a while and finally someone took a deeper look!

chelmuth:
> However, the return code '1' suggest more writing, however that doen't happen.

In the case of `1` above, `connection_handle_write` sets `is_writable` to 0. This hints we may have an issue with our file descriptor handling on poll/select. The code in src/connections.c:1255 calls write only after `is_writeable` was set to one (maybe in `connection_handle_fdevent`).

skalk:
This issue is no longer reproducible.

PR #1839, opened 2015-12-31
Upgrade lighttpd to 1.4.38

gwitmond:
Add hack to serve files bigger than 256kB.
See: https://github.com/genodelabs/genode/issues/987


cnuke:
Thanks for bringing the problem back to our attention. Since this interim fix merely masks an underlying issue with our libc other than addressing it we would rather not to merge it. Upgrading lighty is fine, however. If you would create two seperate commits we could include the updated version.

chelmuth:
@gwitmond I merged the update with fixups for patch removal into staging.

PR #1857, opened 2016-01-20
Use new FSF address in GPL notices

jserv:


nfeske:
Thanks for your interest in improving Genode. However, I fail to see the problem that you are trying to fix with your pull request.

First, I doubt that anyone would take Genode's `elf.h` file to look up the postal address of the FSF. The mere fact that the address is included in the file may be questionable. But adjusting the address seems pointless to me.

Second, the other files modified by your patch are 3rd-party code not maintained by Genode. In particular `repos/ports/src/app/dosbox/config.h` is generated by a configure script, and `repos/libports/src/lib/fuse-ext2/patches/op_access.c.patch` patches 3rd-party source code. In the latter case, your modification actually renders the patch ineffective as it can no longer be successfully applied.

For the reasons stated above, I am not going to merge you patch. I kindly ask you to close the pull request.

jserv:
Defer to the suggestion of @nfeske

PR #1868, opened 2016-01-29
fix the argument type

gapry:
change format form `%lu` to `%u`, since argument has type ` ‘Genode::size_t {aka unsigned int}’`

### DIFF
--- a/repos/os/src/test/vfs_stress/main.cc
+++ b/repos/os/src/test/vfs_stress/main.cc
@@ -510,7 +510,7 @@ int main()
 		elapsed_ms = timer.elapsed_ms();
 
 		for (size_t i = 0; i < thread_count; ++i) {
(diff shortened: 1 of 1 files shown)


chelmuth:
Thanks for your take on these compiler warnings but I fear we're not there yet. Unfortunately, the (unsigned) integer type is always 32-bit while `size_t` can be 32-bit or 64-bit according to the target platform.  For this reason, there exists a special length modifier for `size_t` which is *z*. Therefore, we should replace `%lu` by `%zu` to prevent compiler warnings.
Additionally, the commit message should reflect the issue fixed not just what was done. I'd suggest something like
```
vfs stress: prevent warnings when logging size_t
```

gapry:
@chelmuth Thank you for your reply

PR #1890, opened 2016-02-19
hw_x86_64_muen: Add support for Message Signaled Interrupts

reet-:
This patch set adds support for MSIs to the `hw_x86_64_muen` platform.

The  first patch enables the ACPI functionality in the platform_drv on `hw_x86_64_muen` and
provides a simple generated XML report as ROM session in order to make the PCI configuration space available.

The base-hw `Irq_session_component` class is  extended with member variables to support MSI mode of operation. Values for MSI are only returned if the private `_is_msi` member is set to true.

The platform-specific MSI code is factored out to the `get_msi_params` function which returns false if either the platform or the device does not support MSI mode of operation.

The Muen Sinfo API is extended to retrieve PCI device information via the  `get_dev_info` function. The provided information is required to calculate MSI address/data register values in remappable format (see Intel VT-d specification, section 5.1.2.2).

The `get_msi_params` function on `hw_x86_64_muen` implements support for MSIs by using information provided via the `Sinfo::get_dev_info` function.

The changes are tested with the autopilot tool  for both `hw_x86_64` and `hw_x86_64_muen`.

### DIFF
--- a/repos/base-hw/src/core/include/irq_session_component.h
+++ b/repos/base-hw/src/core/include/irq_session_component.h
@@ -31,6 +31,8 @@ class Genode::Irq_session_component : public Rpc_object<Irq_session>,
 		unsigned         _irq_number;
 		Range_allocator *_irq_alloc;
 		Genode::uint8_t  _kernel_object[sizeof(Kernel::User_irq)];
+		bool             _is_msi;
+		addr_t           _address, _value;
 
 		Signal_context_capability _sig_cap;
 
@@ -58,8 +60,16 @@ class Genode::Irq_session_component : public Rpc_object<Irq_session>,
 
 		void ack_irq() override;
 		void sigh(Signal_context_capability) override;
-		Info info() override {
-			return { .type = Genode::Irq_session::Info::Type::INVALID }; }
+
+		Info info() override
+		{
+			if (!_is_msi) {
+				return { .type = Info::Type::INVALID };
+			}
+			return { .type    = Info::Type::MSI,
+			         .address = _address,
+			         .value   = _value };
+		}
 };
 
 #endif /* _INCLUDE__IRQ_SESSION_COMPONENT_H_ */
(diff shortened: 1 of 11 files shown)


m-stein:
@reet- 
Thank you for these clean commits! I have only some minor suggestions on this. First, are the three fixups that I uploaded OK for you? Additionally, I would suggest to avoid bit logic like it is used by `to_sid` and `to_msi_addr` in `x86_64/muen/platform_support.cc` via the means of the Genode Register/MMIO framework (`util/register.h`, `util/mmio.h`). What do you think?

reet-:
@m-stein
The fixups are certainly fine, thanks for that! I will squash them into my branch and then rewrite the `to_sid` and `to_msi_addr` functions to use the Register/MMIO framework if this is ok for you.

reet-:
@m-stein 
As no registers or MMIO regions are involved, I fail to see how the current `to_sid`/`to_msi_addr` code can be simplified by using the Register/MMIO framework. Could you please provide some hints or even better, a concrete code example?

chelmuth:
@m-stein may we merge this line of work into staging before the release?

m-stein:
@reet- I've added a fixup that illustrates the use of the Register framework. What do you think about it? Could you please test it? It's a little bit more bureaucracy to do it this way but at the other hand it would be more descriptive and type safe I think.

I have not found any names for the bitfields in the VT specs, thus `Bits_0`/`Bits_1`. But I found that handle[15] shall be mapped to MSI-address[2] whereas the previous implementation `... | (handle & 0x8000) << 2;` looked to me as if it is mapped to MSI-address[17]. The Register implementation fixes that. Did I miss a thing here?

reet-:
@m-stein Thanks for the fixup illustrating the use of the Register framework as a replacement for bit logic operations. I think it is not actually easier to read, but the fact that you found a bit logic error in the previous implementation shows that the solution is cleaner.

The relevant autopilot tests pass.

m-stein:
@reet- Thank you for giving it a try :)

chelmuth:
Merged the commits to staging just now.

m-stein:
@chelmuth Added a fixup for RISCV.

PR #1899, opened 2016-03-01
Added rust support

waylon531:
This isn't incredibly useful right now because you can't interface C++ with Rust very well, but you can currently reexport C++ functions with the C ABI.


nfeske:
Using rust on top of Genode sounds indeed like a good idea. Thanks for your initiative!

I just reviewed your patch and like to give the following remarks:

* You introduced a `base-rust` repository. I assume that your intention was to cleanly separate the rust-specific code from the rest of Genode, which is a very good intent. However, the `base-<platform>` repositories serve a different role. Each of those repositories stands for a different kernel (like linux, okl4, sel4, nova etc.). The support for rust is a feature that works across kernels. So the principle support should naturally be hosted in `base` and the rust-library support should be hosted in `libports` (since the library is a 3rd-party library). We may follow the pattern used for the C++ standard library called `stdcxx` and hosted in `libports`.

* On Genode, there is a single name space for library names. Hence, naming your libraries `liballoc`, `libcollections`, etc. is a bit unfortunate because those names are too generic. Instead, they should be prefixed so that it becomes immediately clear that they belong to rust. I.e., I would favor `rust-alloc`. `rust-collections`, etc. (we can leave out the "lib" prefix because we are solely talking about libraries anyway).

* For integrating 3rd-party code with Genode, we have a mechanism called "ports", which is described in Section 5.2. "Integration of 3rd-party software" of the manual (http://genode.org/documentation/genode-foundations-15-05.pdf). Instead of employing the git-submodules mechanism, 3rd-party code should be integrated with Genode's ports mechanism.

* I am not quite sure what to think of the addition of the `target.json` file to `<builddir>/etc/`. Is the user supposed to manually edit this file? If not, we should consider pulling it from a location within the source tree.

* The changes of the build system look good to me. Maybe we find a way to eliminate the nested `ifndef` in global.mk but that is just a minor gripe. ;-)

I assume that the rust support expects an LLVM tool chain to be installed on the host environment. To me, this raises the question about how compatible is this tool chain with the Genode platforms. I.e., our experience with gcc tells us that the compiler will generate code that performs TLS lookups and expects the Linux TLS mechanism to be present (using the gs register on x86_32). On Genode, however, this particular mechanism does not exist. Therefore, we are forced to build a tool chain specifically for Genode. I am curious, why don't you observe such problems with using the host tool chain for rust?

ehmry:
Perhaps we could make an extension system for the makefiles like base/mk/generic.mk that would include makefiles from additional repositories? I bring it up because I also had to make similar changes for Nim.

waylon531:
Rust requires its libraries to be in the format <libname>-[hash].rlib, so I was unable to prefix rust libraries with rust-. However, rust doesn't care what you put in the hash field so I stuck -rust there. I made the target.json file be linked to <builddir>/ instead of <builddir>/etc as the user shouldn't need to edit it. However, they are still configuration files and will likely need to be edited as features are added to Genode. 

I think the reason there are less problems with llvm is because it's designed for cross-compilation. ELF TLS can be turned on or off in the target.json file with has_elf_tls.

nfeske:
Thank you for the adjustments!

I am afraid you slightly mistook my remark about the location of the `target.json` file. I would not host it within the build directory because those files seem to be merely architecture-dependent but not build-directory-specific. Also, they actually do not seem to be strictly rust-related but LLVM-related. So how about placing those files at `repos/base/src/lib/llvm/spec/<arch>/target.json` where `<arch>` is `x86_32`, `x86_64`, `arm`, and `riscv`? Now the only open question is how to point `CC_RUSTC_OPT` to the right file? This can be done by introducing a `repos/lib/import/import-libcore-rust.mk` file. This file will be incorporated each time, a target uses the `libcore-rust` library. Within this file, we can extend the `CC_RUST_OPT` variable depending on the architecture. For an example of this approach, please have a look at `repos/libports/lib/import/import-libc.mk`. Those changes will alleviate the need for any modifications within `tool/`.

The rust-example is pretty empty. How do you think about moving the example to `repos/base/src/test/rust/` instead of a dedicated repository? Ideally, the `config` template could be turned into a proper run script at `repos/base/run/rust.run` (taking cues from `base/run/printf.run`). This way, we will eventually become able to integrate the rust test into our list of automated tests that are executed each night.

nfeske:
Merged to staging

chelmuth:
The commit above changes the download URL from http to https. Last night our buildbot's wget version 3.14 failed to fetch the port sources and the `--no-check-certificate` option did not help too. To fix the issue on our end, i.e., update to newer wget / Ubuntu release, I suggest an internal discussion.

nfeske:
The buildbot stumbled over the attempt to build libbacktrace on hw_pbxa9:

```
contrib/rust-98005b2246d288014ec9d3962e0ec2cd7c9446d0/src/lib/rust/src/libbacktrace/atomic.c:38:33: fatal error: backtrace-supported.h: No such file or directory
```

Commit https://github.com/genodelabs/genode/commit/a7dd14315ce899bb22babc7836aaf81f42b588e7 introduces the the port but the library actually remains unused by the run/rust.run example. In order to reach a consistent intermediate state that could be merged to master, I would like to remove libbacktrace from the commit above. The port of the library could then be added as a separate commit, preferably using the libbacktrace code from the actual upstream source instead of the version that comes with the rust port. @waylon531 do you agree?

nfeske:
I just realized that libbacktrace is indeed used by libstd-rust.mk, but this library remains unused. How should we go about it?

1) Remove a7dd143 from staging for now as it introduces functionality that is not covered by a test.
2) Let the rust test actually use the libstd-rust library and fix the problem of building libbacktrace.

I would actually prefer to do both. In order to avoid commit a7dd143 to block us from merging staging to master, I would temporarily remove it. Once the std-library support is in a state where it is tested, we will include it separately.

nfeske:
With commit 8e205885a8aec20b19f375ffb844fc3541a9875c I tried to untangle the situation. It includes the important bits about the ARM support. I have not included the change to ARM v7 though because this change breaks the test on the Raspberry Pi. Here, we would eventually need a distinction at a finer granularity. Let's see if this commit make our builds and tests happy. @waylon531 are you fine with building the std-library support on top of this commit?

waylon531:
Commit e0059c8 should fix the issues with the raspberry pi, but I haven't been able to run the tests myself as my version of qemu isn't supported. I also removed the Rust standard library target (and libbacktrace) because it doesn't compile.

nfeske:
I actually fixed the Raspberry-Pi issue already while the patch was on our staging branch by leaving out the standard-library-specific parts and removing the ARMv7-specific tool-chain parameters. In the meanwhile, the test runs happily on all platforms and kernels. I have actually added it to our daily tests. The patch eventually reached the master branch: See 28117fee12e84aac5b95e371295b8498037fb7c2

Could you please rebase your latest state on top of the master branch? I think that the distinction between ARMv6 and ARMv7 is much preferable to my solution.

May we take the libstd-rust support to a separate issue and close this one?

waylon531:
Moving libstd to another issue is a good idea. The memory manager also needs some improvement. Rust wants types to have specific alignments. Right now the memory manager just uses malloc, which doesn't let you specify alignment. I saw the allocator for C++, but I'm not sure how to make an interface from that that Rust can use. Another solution would be writing a memory allocator in Rust.

nfeske:
I think it would be fairly straight forward to wrap a (singleton) instance of Genode's `Allocator_avl` behind a C function interface, which can then be called by the Rust code. The `Genode::Heap` may serve as a blue print. Following this route, we would hide the details of the backing-store allocation (allocating RAM dataspaces and attaching them into the local address space) from Rust. Once we have made those parts of the Genode API available to Rust programs, moving the allocator to Rust sounds very attractive.

waylon531:
Commits 3a16ed6 and ea5c0d8 contain fixes that were made after the commit that you merged with master (I also merged the armv6 change into ea5c0d8). They probably should be merged with staging.

nfeske:
Commit https://github.com/genodelabs/genode/commit/3a16ed6698ec831a13261286fbe615f15bedd0ca contains parts that are subsequently removed by commit https://github.com/genodelabs/genode/commit/ea5c0d89d473158a8df6f007c57ade81e60ba42e, in particular libbacktrace and libstd-rust. Wouldn't it be better to clean the commits from these artifacts before merging them to staging?

waylon531:
I merged the two commits together because most of ea5c0d8 was removing things from 3a16ed6.

nfeske:
Thank you for putting both commits into a single one. I merged the result to our staging branch.

Sorry for the late response. We are in the middle of a huge API cleanup phase that keeps my attention captured. Please bear with me. ;-)

chelmuth:
I moved *test/rust* and *lib/rust-targets* to libports with a71fcae. Do you confirm or should rust-targets stay in base?

nfeske:
That is a more suitable location indeed, as the rust support depends on libraries provided by libports.

nfeske:
@waylon531 I tried to reach you via email but got an error "The email address...- is not
    available here.  Disconnecting...

So I am posting my question here: I am currently working on the release notes of the upcoming version of Genode. Of course, I will mention the support for Rust that you added. Would you be fine with me naming you as the author of this contribution?

waylon531:
@nfeske I'm fine with you naming me the author of the contribution.

> I tried to reach you via email but got an error "The email address...- is not
available here. Disconnecting...
I'm not sure why this happened, my mail server is still up and I've still been getting mail from other people.

nfeske:
@waylon531 here is my suggestion for featuring Rust in the release notes: https://github.com/nfeske/genode/commit/2957e9baac7f87965893259ddf1943a255e5db15

Do you approve? Maybe we could add a pointer to the known-to-work rust tool chain? Or are there any technical aspects you'd like to highlight?

waylon531:
Your summary looks good. 

> Maybe we could add a pointer to the known-to-work rust tool chain?

The port uses the nightly rust from 2016-03-03, so the nightly compiler from that day is guaranteed to work. It can be downloaded with `curl -sSf https://static.rust-lang.org/rustup.sh | sh -s -- --channel=nightly --date=2016-03-03` or manually downloaded from https://static.rust-lang.org/dist/2016-03-03/index.html.

nfeske:
Thanks. I added the annotation about the tool chain.

alex-ab:
Merged to master.

PR #1913, opened 2016-03-18
tcp_terminal: disconnect socket if there was no data

ValiValpas:
The tcp_terminal kept the socket open even though the client disconnected. As a result, reconnecting to the tcp_terminal was not working.


nfeske:
Merged to staging (with a slightly adjusted commit message). @ValiValpas do you approve?

ValiValpas:
Sure.

PR #1914, opened 2016-03-18
config: fix issue where config is loaded from an invalid dataspace cap

ValiValpas:
In scenarios where the config ROM is loaded from a report ROM or any other non-static ROM, config might try to access an invalid dataspace capability.


nfeske:
Merged to staging. Again, with slightly changed commit message. Thanks for the fix!

PR #1915, opened 2016-03-18
Ipv4_packet improvements

ValiValpas:
Adding write accessors to the Ipv4 header fields, including checksum generation. Also fixing the header_length field.

### DIFF
--- a/repos/os/include/net/ipv4.h
+++ b/repos/os/include/net/ipv4.h
@@ -66,8 +66,8 @@ class Net::Ipv4_packet
 		 ** IPv4 header fields **
 		 ************************/
 
-		unsigned         _version         : 4;
 		unsigned         _header_length   : 4;
+		unsigned         _version         : 4;
 		Genode::uint8_t  _diff_service;
 		Genode::uint16_t _total_length;
 		Genode::uint16_t _identification;
@@ -132,7 +132,7 @@ class Net::Ipv4_packet
 		 *******************************/
 
 		Genode::size_t  version()     { return _version;                   }
-		Genode::size_t  header_length() { return _header_length / 4;       }
+		Genode::size_t  header_length() { return _header_length;           }
 		Genode::uint8_t precedence()  { return _diff_service & PRECEDENCE; }
 
 		bool low_delay()              { return _diff_service & DELAY;      }
--- a/repos/os/include/net/ipv4.h
+++ b/repos/os/include/net/ipv4.h
@@ -60,6 +60,8 @@ class Net::Ipv4_packet
 
 		static Ipv4_address ip_from_string(const char *ip);
 
+		static Genode::uint16_t calculate_checksum(const Ipv4_packet &packet);
+
 	private:
 
 		/************************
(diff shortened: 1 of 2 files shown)


repos/os/include/net/ipv4.h
```diff
@@ -66,8 +66,8 @@ class Net::Ipv4_packet
 		 ** IPv4 header fields **
 		 ************************/
 
-		unsigned         _version         : 4;
 		unsigned         _header_length   : 4;
+		unsigned         _version         : 4;
```
skalk: why do you change the order of version and IHL header fields here? The standard says that the version field is followed by the header length field, isn't it?

repos/os/include/net/ipv4.h
```diff
@@ -132,7 +132,7 @@ class Net::Ipv4_packet
 		 *******************************/
 
 		Genode::size_t  version()     { return _version;                   }
-		Genode::size_t  header_length() { return _header_length / 4;       }
+		Genode::size_t  header_length() { return _header_length;           }
```
skalk: Obviously, using division is wrong here (I wonder why it ever worked ;-)), but the history here is that the function shall return bytes and not 32bit-word count. So theoretically multiplication by four should do the right thing.

repos/os/include/net/ipv4.h
ValiValpas: I think this was a byte-order issue. Apparently, bitfields are affected by the endianess so both fields where shuffled in the memory layout. I tested this with wireshark.

chelmuth: A strong sign to replace the bit fields with our register/bit set framework.

skalk: Ok, the problem is that although it is one byte (version and IHL) and byte-order is correct here, within bitfields the bit-order is not correctly reproduced here. As an interims fix the re-order is a good solution. In general, I agree to chelmuth that we should get rid of bitfields here and use the register framework (which did not exist yet when those headers got written).
Apart from that, it would be nice when the header length return value would be expressed in bytes instead of 32bit words.

chelmuth:
I'd like to merge these improvements, but am uncertain about the state of the header-length return value discussion... @skalk? @ValiValpas?

skalk:
@ValiValpas @chelmuth as I already wrote, the re-ordering of the bitfields is good as an interims solution before we replace them with the register framework. On the other hand I'd like to replace the header length return value to not return 32-bit word count, but byte count. Because although the IPv4 standard says that this field in the header returns 32-bit fields, I assume most programmers suppose to get byte count when having a size_t return value, and otherwise have to turn the value into byte count anyway.

chelmuth:
@ValiValpas can you confirm that de0a115 and 4b98b6d restore the original function?

ValiValpas:
@chelmuth I couldn't test this in my original setup because the nic_drv for linux is segfaulting. I was able to run the essential part on hw_zynq though and had a look at the packets with wireshark. Looks perfekt!

chelmuth:
Thanks. Regarding nic_drv segfault I assume you're using a recent Linux distro. We had some issues with Ubuntu 15.10 / 16.04 recently and will address these in the next weeks.

ValiValpas:
Yes, I'm using Archlinux:

`segfault at 1e42c18 ip 00007f3d494cd3eb sp 00007ffce0e5fc10 error 4 in libc-2.23.so[7f3d49454000+198000]`

PR #1916, opened 2016-03-18
base: remove duplicates from dependency list by sorting to prevent ar…

ValiValpas:
…gument list explosion


nfeske:
Merged to staging with a slightly adjusted commit message. I shortened the title to fit into the 50-character limit and annotated the commit message with "fixed #<issue number>". The latter records the relationship to the corresponding issue and closes the issue automatically once the patch is merged to master.

PR #1917, opened 2016-03-18
add rom_fs component which translates a ROM into a File_system session.

ValiValpas:


nfeske:
Very nice, including a README and run script - who could ask for more. ;-) The only thing I am a bit uncertain about is the name. Following the naming scheme of our existing servers that indicate the type of the service as the last part of the name, the name `rom_fs` suggests that the component is a file-system server, which it isn't. To make the purpose of the component more obvious, what do you think about renaming it to `rom_to_file`? Or maybe, you have another idea?

ValiValpas:
I agree on renaming it, but I don't have another idea. Or, maybe something like `rom_filer` in the style of `rom_logger`.

nfeske:
Unfortunately `rom_logger` also comes with misleading connotations. It may wrongly refer to Genode's LOG session interface or may indicate that data is appended (as usually for logs). Maybe `rom_writer` would be less ambiguous?

cproc:
> To make the purpose of the component more obvious, what do you think about renaming it to
> rom_to_file?

I like the idea of naming components which convert interfaces similar to hardware interface converters. For example, there exist VGA-to-HDMI converters, which have a VGA interface as input and an HDMI interface as output. So, 'rom_to_filesystem' or 'rom_to_fs' or 'rom_to_file' would look good to me.

nfeske:
Can we agree on `rom_to_file` to move forward?

PR #1926, opened 2016-03-24
hw_zynq: refactor specs

ValiValpas:
This is a generalisation approach of the hw_zynq target. As the boards
typically use UART1 instead of UART0 (used by qemu), we add a spec for
this.
Moreover, hw_zynq does not imply zynq_qemu anymore, so that the support
of particular boards can be placed in third-party or community
repositories (e.g. Genode world).


chelmuth:
I tend to cherry-pick  6fbcb66. @skalk would you agree?

skalk:
@chelmuth: I've reviewed the commit, and prepared an alternative approach with commit https://github.com/skalk/genode/commit/ebf0368eddc0a08810e1cd96505dc4ab65f134e7

@ValiValpas: please, do not feel offended. I really appreciate your work, and do not want to put obstacles in your way. On the other hand I have to take care to hold the base-hw repository as clean and clear as possible. As long as we do not support other zynq platforms than Qemu officially, I would like to keep the additional definitions (like serial using UART1) away from the official repository. Moreover, I have removed the additional SPEC variables for zynq_uart*, and instead put the needed explicit include path into the library definition for the already existent zynq_qemu SPEC variable. If you do the same for zynq_parallela etc. you can put those together with the definitions I've left away into a follow up commit for genode-world. Are you ok with this?

ValiValpas:
@skalk: I understand your concern and, in general, I'm okay with that (although I haven't adopted and tested it yet).

However, there are two "issues" that come to my mind:
1. If you use the Xilinx version of Qemu, you can actually use the UART1 as you pass a dtb file to qemu. We're not using this version though, but maybe this is something you want to keep in mind.
2. How is your approach w.r.t. the Board_base specification? I see that you removed the additional definitions for the UART1, SD and TTC. Do you think these definitions should go into the world repository as well? I guess this could be quite confusing for a lot of users if these definitions are spread across a lot of files and repositories.

skalk:
@ValiValpas : regarding the two issues:
1. I removed UART1 from the core-only regions. When you want to use UART1 on zynq_qemu you could only use it within core otherwise. Alternatively, we can add "UART_KERNEL = " in the most specific board description file, and use that enum for the core-only regions.
2. First and foremost I did not integrate the additional I/O memory definitions in the commit, because they did not fit to the commit message. Apart from that, I would like to add such definitions where they are actually used. As long as e.g.: the sd card driver is not part of the mainline Genode repository, its I/O memory descriptions should be part of the corresponding repository/branch. If you like to add an additional sub-repository in genode-world that contains ports for other Zynq platforms, you will have to add something like ``include/spec/zynq_*/drivers/board_base.h`` anyway. Such a board description can imply sd-card controller definitions as well.

ValiValpas:
@skalk : Thanks for the clarification. I'm okay with that. I actually like the UART_KERNEL approach very much.

skalk:
@ValiValpas @chelmuth : I've changed the commit according to the discussion around "KERNEL_UART": https://github.com/skalk/genode/commit/5251019cf7cfa0ca66be929f32369220d10f8138. From my point of view, it's ready for staging.

chelmuth:
Merged just now.

PR #1943, opened 2016-04-20
Minor refactoring & dde_bsd port fix

Kensan:
Two changes for your consideration:
- hw_x86_64_muen: Use Bitset template for MSI address handle
- Update dde_bsd port


chelmuth:
Merged both commits into staging.

PR #1953, opened 2016-05-03
base-hw: fix SCU initialization for zynq-based boards

ValiValpas:
The old implementation cleared all other bits in the SCU control
register when enabling the SCU, which broke the kernel startup on zynq-
based boards.
By only raising the enable bit, we can keep the initial/default state
e.g. as set up by uboot.


skalk:
@ValiValpas thanks for the patch!

PR #1955, opened 2016-05-04
base-hw: refactor core.mk and core.inc files

ValiValpas:
All core.inc files now use $BASE_HW_DIR instead of $REP_DIR. The former
is defined by the core.mk file. This allows including core.inc files
from other repositories (e.g. genode-world) for additional platform
support.

See https://sourceforge.net/p/genode/mailman/message/35060584/

### DIFF
--- a/repos/base-hw/lib/mk/core.inc
+++ b/repos/base-hw/lib/mk/core.inc
@@ -11,8 +11,8 @@ LIBS += core-perf_counter
 LIBS += base-common
 
 # add include paths
-INC_DIR += $(REP_DIR)/src/core/include $(BASE_DIR)/src/core/include
-INC_DIR += $(REP_DIR)/src/include      $(BASE_DIR)/src/include
+INC_DIR += $(BASE_HW_DIR)/src/core/include $(BASE_DIR)/src/core/include
+INC_DIR += $(BASE_HW_DIR)/src/include      $(BASE_DIR)/src/include
 
 # add C++ sources
 SRC_CC += console.cc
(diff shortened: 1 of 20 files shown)


skalk:
@ValiValpas thanks for implementing the idea!

chelmuth:
@ValiValpas @skalk I added 75c5fa2 to staging but also had to fixup with f1e7ac7. Do you both confirm?

ValiValpas:
@chelmuth sure

chelmuth:
Tonight, we got another build error
```
.../repos/base-hw/lib/mk/spec/cortex_a8/core.inc:18: /lib/mk/spec/arm_v7/core.inc: No such file or directory
```
and I was willing to fix the same way as with hw_x86_64. On second thought, I had a thorough look at your initial patch and what puzzles me are lines like
```diff
-INC_DIR += $(REP_DIR)/src/core/include/spec/arm_v7
+INC_DIR += $(BASE_HW_DIR)/src/core/include/spec/arm_v7
```
in *repos/base-hw/lib/mk/spec/arm_v7/core.inc*. If all places where `REP_DIR` was used are replaced by `BASE_HW_DIR`, which may point to a special directory outside of the *base-hw* repo, I expect you have to mirror the include hierarchy. Please correct me if I'm wrong. Do you have a example repo containing your additional  hw_zynq specs?

ValiValpas:
@chelmuth The idea was that, unlike `REP_DIR`, `BASE_HW_DIR` is supposed to always point to the base-hw repo. The build error is caused by the fact that I missed the core-trustzone files, hence fixing this the same way as you did for core-muen should work.
Nevertheless, I can point you to the my branch of the world repo which cointains the additional hw_zynq specs:
https://github.com/ValiValpas/genode-world/tree/zynq-boards

chelmuth:
Seems like I mixed things up. Thanks for the explanation.

skalk:
After thinking about above issues, and the whole changeset including adding ``Base_HW_DIR`` in all core.mk files: I would suggest to replace the ``REP_DIR`` usages in core.inc files of the base-hw directory with ``GENODE_DIR/repos/base-hw`` instead. Thereby we can pass the additional declaration.
@nfeske: as you initially gave the inspiration for this change, do you have an opinion?

nfeske:
@skalk I like your idea much more than my original suggestion.

skalk:
@chelmuth: please do not forget to merge commit 957bbcdf3dc3dfc0c0a84792fc70010de5dbcae1 to staging

chelmuth:
@skalk thanks for the reminder - merged it just now.

PR #1992, opened 2016-06-01
Miscellaneous Virtualbox fixes

Kensan:
A couple of minor Virtualbox-related fixes.

### DIFF
--- a/repos/ports/run/vbox_win.inc
+++ b/repos/ports/run/vbox_win.inc
@@ -74,7 +74,7 @@ append config_of_app {
 		needs to be enabled in the .vbox file (disabled by default).
 		(Recording is configured to use the external mic.)
 	-->
-	<start name="audio_drv">
+	<start name="audio_drv" priority="-1">
 		<resource name="RAM" quantum="9M"/>
 		<provides>
 			<service name="Audio_out"/>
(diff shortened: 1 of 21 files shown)


repos/ports/run/vbox_win.inc
```diff
@@ -74,7 +74,7 @@ append config_of_app {
 		needs to be enabled in the .vbox file (disabled by default).
 		(Recording is configured to use the external mic.)
 	-->
-	<start name="audio_drv">
+	<start name="audio_drv" priority="-1">
```
chelmuth: I'm surprised that a priority downgrade of the audio driver eliminates choppy sound. Did you investigate the interactions of the scenario?

chelmuth:
I merged the PR but skipped 9e413148b520f61ed0be05da60c580d304a89f78 for now.

repos/ports/run/vbox_win.inc
Kensan: The choppyness is not eliminated completely but the quality of the output is improved. The issue was discovered when testing audio on hw_x86_64_muen/Virtualbox and @cnuke suggested the change until he has time to investigate the underlying cause. Maybe @cnuke can comment if it still should be applied.

cnuke:
9e41314 should be applied because all other drivers are also running on priority `-1`and it was an oversight on my side that the audio_drv is not. There is no reason that it runs on the highest priority and in fact it also makes the sound choppy when running on NOVA. I still have not investigated why changing the priority of the driver influences the playback quality.

chelmuth:
Thanks, 9e413148b520f61ed0be05da60c580d304a89f78 was merged just now.

chelmuth:
Merged into master.

PR #1994, opened 2016-06-02
Make Muen sinfo API available to userspace components

reet-:
This patch set moves the Muen sinfo API from base-hw _core_ to _base-common_ to make it accessible to userspace components. This is in preparation for the upcoming support of vbox in hw_x86_64_muen.

The implementation is changed from the static functions approach to class instances with the base address of the sinfo memory region in the constructor. The sinfo region is exported as ROM module in the new platform `_init_additional()` function.

This allows userspace components to attach a ROM dataspace and construct an sinfo instance with the address returned by the `rm_session()->attach()` function.

The timer and platform support code uses a common `sinfo()` function to instantiate and access the sinfo object in the kernel.

### DIFF
--- a/repos/base-hw/src/core/include/spec/x86_64/muen/timer.h
+++ b/repos/base-hw/src/core/include/spec/x86_64/muen/timer.h
@@ -14,12 +14,13 @@
 #ifndef _CORE__INCLUDE__SPEC__X86_64__MUEN__TIMER_H_
 #define _CORE__INCLUDE__SPEC__X86_64__MUEN__TIMER_H_
 
+/* base includes */
 #include <base/printf.h>
 #include <kernel/types.h>
+#include <muen/sinfo.h>
 
 /* core includes */
 #include <board.h>
-#include <sinfo.h>
 
 namespace Genode
 {
--- a/repos/base-hw/src/core/include/spec/x86_64/muen/timer.h
+++ b/repos/base-hw/src/core/include/spec/x86_64/muen/timer.h
@@ -63,6 +63,10 @@ class Genode::Timer
 		Timer() :
 			_tics_per_ms(sinfo()->get_tsc_khz())
 		{
+
+			/* first sinfo instance, output status */
+			sinfo()->log_status();
+
 			struct Sinfo::Memregion_info region;
 			if (!sinfo()->get_memregion_info("timer", &region)) {
 				PERR("muen-timer: Unable to retrieve time memory region");
--- a/repos/base-hw/src/core/include/spec/x86_64/muen/timer.h
+++ b/repos/base-hw/src/core/include/spec/x86_64/muen/timer.h
@@ -17,10 +17,10 @@
 /* base includes */
 #include <base/printf.h>
 #include <kernel/types.h>
-#include <muen/sinfo.h>
 
 /* core includes */
 #include <board.h>
+#include <sinfo_instance.h>
 
 namespace Genode
 {
@@ -37,8 +37,6 @@ class Genode::Timer
 		using time_t = Kernel::time_t;
 
 		enum { TIMER_DISABLED = ~0ULL };
-
-		Sinfo _sinfo;
 
 		uint64_t _tics_per_ms;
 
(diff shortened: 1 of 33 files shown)


repos/base-hw/src/core/spec/x86_64/muen/platform_support.cc
```diff
@@ -100,11 +100,22 @@ bool Platform::get_msi_params(const addr_t mmconf, addr_t &address,
...
+
+	if (!result.size) {
+		struct Sinfo::Memregion_info region;
+		if (!sinfo()->get_memregion_info("ram", &region)) {
+			PERR("Unable to retrieve base-hw ram region");
+			return 0;
+		}
+
+		result = { .base = region.address, .size = region.size };
+		PDBG("RAM base 0x%lx, size 0x%lx", result.base, result.size);
```
skalk: Nowadays, we try to circumvent the usage of the PDBG macro in production code. The macro is meant for debugging only, and should vanish when entering staging/master branches. If you really need the printed information, please use Genode::log() instead.

skalk: In fact we can leave it out here and instead enable "verbose" flag in repos/base-hw/src/core/platform.cc, which prints out that information too. The "verbose" flag will be replaced soon, so that the information gets displayed everytime.

repos/base-hw/src/core/include/spec/x86_64/board.h
```diff
@@ -34,4 +34,4 @@ namespace Genode
 	};
 }
 
-#endif /* _CORE__INCLUDE__SPEC__X86__BOARD_H_ */
```
skalk: Just for clarification: are these definitions now 64bit-specific?

skalk: As addition: we can of course get rid of the distinction between x86 and x86_64 by moving everything to x86_64. As long as we do not support x86_32 in base-hw (hopefully never ;-)) this is a rational reason from my point of view. But than we should to this consistently and in an extra issue. Do you agree?

m-stein: :+1: for removal of the ``x86`` (without ``_64``) spec.

reet-: Moving x86/board.h to x86_64/board.h is not necessary atm, I will revert it. The constants are not 64bit-specific of course.

If you want to get rid of the x86/x86_64 distinction really depends on your roadmap for base-hw. But I agree that this is definitely a separate issue.

repos/base-hw/src/core/spec/x86_64/muen/platform_support.cc
reet-: Understood, will drop the PDBG here.

skalk:
@reet: thank you for following my nitpicking comments!
@chelmuth: I've reviewed the pull-request, and I'm fine with it. I would however squash them together into one commit like: "base-hw: make Muen sinfo API publicly available"

repos/base-hw/src/core/include/spec/x86_64/muen/timer.h
```diff
@@ -63,6 +63,10 @@ class Genode::Timer
 		Timer() :
 			_tics_per_ms(sinfo()->get_tsc_khz())
 		{
+
+			/* first sinfo instance, output status */
+			sinfo()->log_status();
+
```
chelmuth: Why not log the status on construction of `Sinfo`?

chelmuth:
I merged the PR. I'm still ready to squash the series and may apply a fixup regarding my comment above.

repos/base-hw/src/core/include/spec/x86_64/muen/timer.h
reet-: I want the memory/channel info logged only once and decided in favor of an explicit call to `log_status()` at the first instance.

chelmuth:
Merged with c1bb82fb6b6f4d7f4aad91c31a5fe9e94caba8b8.

PR #2008, opened 2016-06-14
base_hw: fix run scripts using core_test = 1

ValiValpas:
Without this fix, e.g. `make run/hw_test` fails with:

> mv: cannot stat 'var/run/hw_info/genode/config': No such file or directory

### DIFF
--- a/tool/run/boot_dir/hw
+++ b/tool/run/boot_dir/hw
@@ -146,7 +146,9 @@ proc run_boot_dir {binaries} {
 		exit -1
 	}
 	clean_boot_modules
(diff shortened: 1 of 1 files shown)


m-stein:
@ValiValpas This is already fixed on current master by adding a dummy config to the core tests. I'd like to stay with the dummy config solution to keep run script behavior as generic as possible. What do you say?

m-stein:
(The commit I mention is 608df38)

chelmuth:
As @m-stein mentioned master is fixed.

PR #2053, opened 2016-08-05
Fix up the generation of invocation.h in base-sel4

fzyz999:
Currently, the generated sel4/invocation.h, sel4/arch/invocation.h, sel4/sel4_arch/invocation.h have same header file safeguard(`__LIBSEL4_INVOCATION_H`). As a result, some definitions are missing when a source file includes these invocation.h. To fix this, we should pass correct parameters to invocation_header_gen.py(`--arch` for sel4/arch/invocation.h and `--sel4_arch` for sel4/sel4_arch/invocation.h).


fzyz999:
Sorry, I should check the staging branch first. This has been fixed.

PR #2055, opened 2016-08-09
base: implement print() functions for floating point types

ValiValpas:


nfeske:
The patch looks very good!

I am still a little bit hesitant to merge it for the following reason: As a matter of principle, the base repository contains only code that is needed by the base and which cannot reside somewhere else. The base code is the part of the framework that fundamentally needs most attention with respect to evaluation (or possibly formal verification in the future). When viewed in this light, each line of code within base is a burde. Hence, I try to keep it as small as possible. Because of this line of reasoning, most higher-level convenience utilities are located in other (less critical) repositories such as os or gems.

The base code does not use floating point numbers (at least, it shouldn't). So there is no use case for the output of this type within base. So strictly speaking, the code should not be included there. On the other hand, `float` and `double` are basic types. From a Genode-user point of view, it would be intuitive to have it supported by default.

Admittedly, I'm struggling between both positions. @chelmuth What do you think?

chelmuth:
I vote for integration for two reasons. First, the patch is small and lean and, second, we already have `inline size_t ascii_to(const char *s, double &result)` in *util/string.h* with the same pragmatic considerations.

nfeske:
@chelmuth then go for it! :-)

chelmuth:
Added to staging.

PR #2080, opened 2016-08-27
Fix missing "#include<heap.h>" in hello tutorial.

sighingnow:
A include declaration of `heap.h` is missing in hello tutorial. The type `Sliced_heap` is used in hello server.


nfeske:
Thank you very much! This problem was introduced by the commit "base: avoid use of deprecated base/printf.h", which is fortunately not on our master branch yet. I added a fixup commit to staging now.

sighingnow:
OK. Now this PR should be closed.

PR #2097, opened 2016-09-13
libc: calloc addr should be checked for possible failed allocations

mennovalkema:
@nfeske @chelmuth


chelmuth:
Merged a slightly changed patch as  1f1e028. @mennovalkema do you agree?

mennovalkema:
@chelmuth : great, this also solves the problem!

Issue #2166, opened 2016-11-07
init: print out reason for session creation error


trimpim:
We have a relative complex scenario with many components and much more routes between them.
It is no trivial task to find out why the creation of a session failed without specific error message.
A short message should be printed so that the different cases can be distinguished.

nfeske:
I don't think that the low-level child.cc code is the appropriate place for these messages. From the child's perspective, this condition is an error. But from the parent's perspective (where the child.cc code resides), this condition does not present a problem.

That said, I agree that printing the diagnostic information is valuable. How do you think of adding the warning to the `Connection` class instead. Please have a look at my commit e63bfab074323a46d8064ca245273ec1c582f796

trimpim:
In  principle I'm not against moving the information to a better location. But it should still be possible to distinguish between the different cases, otherwise we will end up with the same situation we have now, that we need to dig in to the code (add error messages) to find out what went wrong.
IMHO it is an error if the connection can't be opened.

ehmry:
I think some guidelines on when and where errors should be printed to the log would be helpful.

nfeske:
I am actually planning to remove the distinction between the `Service_denied` and `Unavailable` exceptions of the `Parent` interface because it should be of no matter to the child to know whether the requested service exists or not. The message that a connection of a certain type could not be established is certainly useful. This is covered by my patch. Furthermore, if init is not able to route the session to a service, it actually prints a message. E.g., try to change the following routing configuration in hello_tutorial/run/hello.run:

```
  <start name="hello_client">
    <resource name="RAM" quantum="1M"/>
    <route>
      <any-service> <parent/> </any-service>
    </route>
  </start>
```

With this configuration, there exists no route to the hello server. When starting the scenario, init prints the following message:

```
[init] Warning: hello_client: no route to service "Hello"
```

Hence, for the `Unavailable` condition, init already prints the diagnostic information. For the `Service_denied` condition, my patch causes the client to print the information:

```
[init -> hello_client] Warning: Hello-session creation failed (ram_quota=4K)
```

Regarding the distinction between warnings and errors, this is a matter of perspective. What may be catastrophic for one component (e.g., hello client aborts because it cannot open a session to the hello server) is a mild inconvenience for another (init noticing that one child is apparently doing "stupid" things like requesting a session it shouldn't access). So maybe it would be sensible to print the message in the `Connection` object as an error?

[...]

PR #2167, opened 2016-11-07
Issue2166

trimpim:
This patch adds the error messages that help us solve session creation problems.


chelmuth:
#2166 was closed by issuer.

PR #2212, opened 2016-12-28
Fixed build on Ubuntu 16.10 for x86_64.

drgraetz:
Ubuntu provides a position independent shared object for libsdl-1.2-dev.
To appropriatly link it to Genode, the linker flag '-no-pie' had to be
added to the make file.

### DIFF
--- a/repos/base-linux/lib/import/import-lx_hybrid.mk
+++ b/repos/base-linux/lib/import/import-lx_hybrid.mk
@@ -109,3 +109,8 @@ LD_CMD = c++
 
 # disable format-string security checks, which prevent non-literal format strings
 CC_OPT += -Wno-format-security
+
+# disable position-independent executables (which are enabled by default on
+# Ubuntu 16.10 or newer)
+CXX_LINK_OPT += -no-pie
+


chelmuth:
I merged the commit with 4faef2658030bc17b500f9a8da05eecc5dc797de. Please note that the email address used for the commit is not linked to your github account currently. This can easily be fixed by following the instructions on the page linked by the commit ID (mind the small (?)).

chelmuth:
That's a pity mostly because we do not use any test for supported command-line parameters currently. What is `c++ --version` on 14.04?

nfeske:
The version is:
```
g++ (Ubuntu 4.8.4-2ubuntu1~14.04.1) 4.8.4
```

Instead of adding version heuristics, however, I would rather employ some probing magic, like this:

```
CXX_LINK_OPT_NO_PIE := $(shell ($(CC) -no-pie 2> /dev/null > /dev/null && echo "-no-pie") || true)
CXX_LINK_OPT += $(CXX_LINK_OPT_NO_PIE)
```

chelmuth:
Yeah, that is what we had years ago before we had our own tool chain ;-) I was just wondering which version of GCC does not know about `-no-pie`. If the code above works for you, I suggest adding a fixup before we enable the buildbot again.

nfeske:
I added the fixup https://github.com/genodelabs/genode/commit/5d1687f9d3dccc1c62050f4c31ab50f097bf758f. It works for Ubuntu 14.04. Could you please give it a try with a version of GCC that supports `-no-pie`? (see if `-no-pie` actually appears as linker argument)

chelmuth:
It does not appear on my command line for the linking stage, but I'll investigate.

chelmuth:
I had to add more GCC flags on the command line and it works now on 16.04 with 04a0b8f912be849e4206daf90422cc133f5ee800. @drgraetz could you please check with Ubuntu 16.10? Just running `make drivers/framebuffer/spec/sdl` successfully will be sufficient.

drgraetz:
I successfully rebuild the entire linux_x86 build on Ubuntu 16.10 using
changeset 04a0b8f.

A happy new year!

chelmuth:
@drgraetz very good, thank you.

chelmuth:
Merged into master with b2934c7aec8e08d4b2c388bba41f78010fd4ae74

PR #2220, opened 2017-01-03
report_rom: add enabled() and enabled_sigh()

ValiValpas:
This allows the reporter to detect whether any ROM session is subscribed
to the report. If not, writing and updating the report can be omitted to
spare some CPU cycles. This is also good for debugging.

### DIFF
--- a/repos/os/include/os/reporter.h
+++ b/repos/os/include/os/reporter.h
@@ -120,6 +120,18 @@ class Genode::Reporter : Noncopyable
 			_conn->report.submit(length);
 		}
 
+		void enabled_sigh(Signal_context_capability cap) {
+			if (_enabled)
+				_conn->report.enabled_sigh(cap);
+		}
+
+		bool reporting_enabled() {
+			if (_enabled)
+				return _conn->report.enabled();
+			else
+				return false;
+		}
+
 		/**
 		 * XML generator targeting a reporter
 		 */
--- a/repos/os/include/report_rom/report_service.h
+++ b/repos/os/include/report_rom/report_service.h
@@ -41,6 +41,8 @@ struct Report::Session_component : Genode::Rpc_object<Session>, Rom::Writer
 
 		Rom::Module &_module;
 
+		Genode::Signal_context_capability _enabled_sigh;
+
 		bool &_verbose;
 
 		Rom::Module &_create_module(Rom::Module::Name const &name)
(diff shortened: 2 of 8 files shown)


nfeske:
That is an interesting suggestion. But I feel that it takes us to a wrong direction, for the following reasons:

* It increases the complexity of the report-session interface and thereby the protocol the report producer has to follow. Let me illustrate an issue: Right now, the report-producer may generate a report in a fire-and-forget fashion, e.g., when a certain event occurs. Thereby, it can easily supplement transient information like the current time when the event occurred. With your change  - if there is no subscriber for the report at the time of reporting - the report is not issued. Now, when a new subscriber appears, the current state is unknown at the report-rom. Hence, the report-client has to re-generate the report as a response to an `enabled` signal. But at this point, the time of the original event is lost. Consequently, the report-client has to remember the time of the last event, just to follow the protocol. I believe that we can address this specific issue case by case. But the mere existence of such corner cases and the need for the attention of the API user is an indicator that we should better avoid the added complexity.

* The mechanism adds a hidden communication channel from the ROM consumer to the report producer, which is not obvious and may contradict with the intention behind a report-rom configuration. Think of a ROM consumer that conspires with the report producer. The report-rom policy dictates that information may flow from the report producer to the ROM consumer but not vice versa. From the information-flow point of view, the report-rom server is a data diode. With your suggestion, this nice property vanishes because the report producer may misuse the ROM sessions to the report-rom server as a covert channel by repeatedly opening/closing the ROM session. In fact, the channel would actually have a high bandwidth because the rate of ROM-session create/close requests is not limited.

* Related to the both points above, the added protocol would create a possible denial-of-service vector from ROM consumers against report producers that generate heavy-weight reports at a low rate. Because of (1), the report-producer must produce such a report each time when `enabled` becomes `true`. This condition can be triggered by any of its ROM consumers by repeatedly creating and closing the ROM session. Because the report-producer performs the report-producing work as a response to an asynchronous notification (as opposed to an RPC call), it has to pay for the processing time needed to produce the report. Consequently, any ROM consumer may artificially put load on the report producer.

There is actually an obvious solution for limiting the rate of reports generated: Let the report producer obtain a ROM module that contains the information when and how reports should be generated. In fact, the component's "config" is such a ROM module. E.g., nitpicker responds to config updates and re-evaluates the `<report>` XML node. But in principle, this information could be provided by a different ROM module - let's say - a "report_rate" ROM. Now, anyone with the ability to provide this ROM is able to steer the behavior of the report producer. In the special case addressed by your pull request where the ROM consumer wants to control the report rate at the report producer, we could use a `report_rom` config as follows:

```
<config>
  <policy label="rom_consumer -> state" report="report_producer -> state"/>
  <policy label="report_producer -> report_rate" report="rom_consumer -> report_rate"/>
</config>
```

Note that in this example, the report-ROM configuration explicitly expresses the consent that information may flow from `rom_consumer` to `report_produce`. But the nice thing is that the rate cannot be controlled only by the ROM consumer (as with your suggestion) but by anyone that generates a "report_rate" report. E.g., you could control the rate of many report producers with just a single global "report_rate" report.

ValiValpas:
@nfeske That's a very convincing argumentation.

PR #2317, opened 2017-03-12
Trace execution time

jobnz:
As part of my masters thesis, i needed the execution time of threads along with any trace event corrresponding to a thread. These two commits implement thread execution time for base-hw and access to thread execution time from Trace API. Since Genode 16.05 access to the Native_utcb, for example, was removed from the public Thread API, which is used in this implementation. Thus, this pull request is not intended to be merged directly into Genode but to discuss how to implement this the Genode-Way.


nfeske:
Thanks for putting your patches up for discussion!

Implementing `Platform_thread::execution_time` is certainly welcome. As a nice side effect, it enables the use of the CPU-load monitor on base-hw (right now, it only works well on NOVA).

I am hesitant about your suggestion to extend the `Thread` interface with an `execution_time` method and your extension of the trace-policy API, for the following reasons:

* The extension of the generic interface suggests that the mechanism works across kernels, which is not the case.

* It adds complexity to the generic API to accommodate a single use case (that I admittedly don't fully understand).

* Introducing new arguments to the trace-policy functions sets an unwelcome precedent for future extensions. E.g., if someone needed to capture the CPU affinity of a given thread in time and followed the footsteps of your patch, the policy functions would be further extended by a CPU-affinity argument. The interface would eventually become more and more complicated.

I presume the use case of the `execution_time` interface is the gathering of holistic information about the execution time and scheduling of threads. To cover this use case, I'd vastly prefer a more powerful mechanism: Letting the kernel's scheduler generate a trace event whenever a scheduling decision is taken. It would roughly work as follows:
* We allow the trace mechanism to trace the kernel (as a trace subject). In principle, this would allow the trace monitor to extract plenty of valuable information from the kernel such as point in time when an interrupt occurs, when scheduling decisions are taken, etc. The trace policy for the kernel would be hardwired (at least at first) but the trace activation would be triggered by the trace monitor.
* We add a trace point to the kernel's scheduler that captures the scheduling decision.
* Both the trace policy in the kernel as well as in the user-level components supplement the trace events with the value returned by `Trace::timestamp()`. This value would be generated by the trace policy internally, alleviating the need to extend the trace-policy interface.

Since the trace events of the kernel's scheduler and the user land are accompanied with a global time stamp, the streams of events can be correlated to infer a global picture of the thread execution times.

This approach has the following advantages:

* It does not need to extend the Genode API nor the trace-policy interface.
* It does not need the active help of the traced components.
* It would yield more accurate results as the kernel's scheduling decisions are captured directly.
* It would be suitable for obtaining any kind of kernel events (page faults, signals, scheduling decisions, interrupts, VM exits). So should the need for tracing such events arises, we already have a principle way to go.

That said, it has the drawback that meaning of the value of `Trace::timestamp` depends on the platform. E.g., on multi-core systems, the values returned by the different CPUs is not guaranteed to be consistent.

@jobnz What do you think of this alternative approach? Would you be able to cover your use case with it?

ValiValpas:
@jobnz thanks for creating the pull request and starting the discussion

@nfeske let me try to fill in:
First of all, I like the idea of kernel tracing very much as it seems way more powerful. I think this should be an alternative approach that completely covers our use cases and might even eliminate a shortcoming that we accepted.

However, let me still explain what motivated our approach:

- The main idea was that this approach is easily extensible to other kernels as long as they perform some accounting of the thread execution time (e.g. NOVA). As the execution time is already part of the TRACE API, it seemed straightforward to use this in the policy interface.
- I think a benefit of our approach is that it implicitly allows to detect whether there was a preemption between two trace events of a thread without the need to trace any other thread (or the kernel) and potentially process a lot of trace/scheduling events to search for the relevant timestamps. (I.e. if two consecutive trace events of my thread have logged the same execution time there was no preemption.)

@jobnz Do you have anythink to add?

jobnz:
@nfeske Thanks for helping!

I agree with both of you. Tracing the kernel is more generic and seems to be more powerfull than the way we implemented it. I also think that our use cases can be covered this way as well.

ValiValpas:
Let me revive the discussion here: I just uploaded a new branch which adds support for adding a trace policy to core (base-hw). At the moment there is only a single core event to keep it clean (I have also experimented with other events though). The thread ids are taken from the kernel-internal addresses of the corresponding objects. A mapping between ids and thread labels is accessible via the TRACE session (separate commit).

nfeske:
Since we don't stress Fiasco.OC beyond nightly tests, nobody seems interested in this feature since 3 years. Hence, I'm closing it.

jschlatow:
Fine for me. But just in case someone will stumble upon this issue in the future: It was not about Fiasco.OC but for base-hw. The initial PR might suggest otherwise, yet the revised solution consisted in adding trace-policy support solely to base-hw. If we ever need this feature at some future point in time, I can update my outdated feature branch and create a new PR.

nfeske:
@jschlatow Thank you for the acknowledgement.

PR #2346, opened 2017-03-23
Fix a socket leak in tcp_terminal when it is closed remotely.

pmundkur:
I noticed this when reading the code.


nfeske:
Thanks for the fix! I merged the patch to genodelabs/staging just now.

Issue #2367, opened 2017-04-05
dde_linux: more generic tar lists.


jklmnn:
The .list files for the tar archives should be more generic. I think it is quite inconvenient to change all paths in these files if a kernel version changes.

jklmnn:
See #2368.

PR #2368, opened 2017-04-05
dde_linux: generic tar lists

jklmnn:
I changed the versioned path with a placeholder (replaced `linux-4.4.3` with `linux-x.x.x`). Instead of using `--files-from` tar not gets the required files from a sed command which replaces the placeholder with the intended version.
For an explanation, see #2367.


ssumpf:
@jklmnn: 75402c0 may overflow the command line length, which is why we started the file lists in the first place, doesn't, e.g.,`:%s/4.4.3/4.x.x/` do the job for your? One only needs do perform this once per Linux version upgrade, which tends to happen every one or two years.

chelmuth:
If this is really desirable (this is not my choice and I depend on @ssumpf and @cnuke for a reasonable judgement) I suggest to stay with the following feature (from the man page)

> If you give a single dash as a file name for `--files-from', (i.e., you specify either --files-from=- or -T -), then the file names are read from standard input.

in combination with `sed`.

BTW, you need not open a separate issue for a pull request as both are handled almost the same on GitHub (at least with our workflow).

alex-ab:
> BTW, you need not open a separate issue for a pull request as both are handled almost the same on GitHub (at least with our workflow).

@jklmnn: use  ``#number`` in your commit messages and links to the commit will appear automatically in the github issue

jklmnn:
@ssumpf I shortened it (but I preserved a leading `-` to prevent matching the last `x` of `linux-`).
@chelmuth I readded `--files-from` and changed the sed command to write on its standard input.

jklmnn:
Is there anything left to do to merge this PR?

chelmuth:
Thanks for the reminder. Please squash your commits into one feature commit and prefix the descriptive commit message with `dde_linux:`. I'll merge as soon as you notify me about the updated commit.

jklmnn:
@chelmuth I have squashed the changes into https://github.com/jklmnn/genode/commit/a3ab57f3c5e9b0bacb0445a4bbdcecb8e9428c42.

Issue #2242, opened 2017-01-18
UEFI x86 support


alex-ab:
will require another boot-loader than the currently used GRUB legacy. Either we have to develop something own special or go with GRUB2, which seems to be the natural choice.

For the moment I experimented a bit with GRUB2 and that are the current findings with loading Genode/NOVA and Genode/hw via GRUB2 UEFI.

Serial/UART output
--------------------------
Up to now we used Bender as a chained bootloader to lookup the PCI devices which provide a Serial/UART interface. Bender wrote the findings about the I/O ports into the Extended BIOS Data Area (EBDA), which the kernels and Genode/core looked up to print messages via the UART.

The question is, whether we can keep this way for UEFI (EBDA is not anymore) or we have to find another way, e.g.
- adding a Multiboot or Multiboot2 element about the I/O ports  (non-complex, convenient, not MULTIBOOT/MULTIBOOT2 standard)
- or we expect that the kernel or Genode/Core have to lookup the PCI devices by itself (complex but convenient)
- or if they have to be configured manually (in-convenient).

For the moment I hardcoded the I/O ports in the kernel and Genode/core, which is really in-convenient. But even more in-convenient is, that the PCI serial card for AMT SOL is by default not usable on my testmachine. One has to manually enable in the Command Register of the PCI device the I/O Space bit (http://wiki.osdev.org/PCI) for each boot - grub2 terminal
```
setpci COMMANDS
setpci -s xx:xx.x 4.b=1
```

ACPI
----
The lookup of the ACPI RSDP (Root System Description Pointer) the old way seems to be not possible anymore. The pointer was located either within the first 1KB of the EBDA or in the main BIOS area 0xE0000 to 0xFFFFF.
In the UEFI case the pointer gets reported via the EFI_SYSTEM_TABLE to the bootloader, which is in our case GRUB2.

Unfortunately, GRUB2 does not report the RSDP pointer via the Multiboot structure, however does so for the Multiboot2 structure.

For that reason I extended NOVA to be able to be loaded as Multiboot2 compliant kernel and to parse the Multiboot2 structure. With this information I was able to boot all CPUs and to start Genode/core. Genode/hw booted as normal Multiboot compliant kernel out-of-the-box, since it does not require (currently) the RSDP ACPI pointer.

[...]

PR #2409, opened 2017-05-10
os: added framebuffer driver that uses the multiboot2 framebuffer

jklmnn:
I implemented a framebuffer service that uses the framebuffer provided by multiboot2. This PR just contains the driver for Genode. Since it requires an adapted hypervisor information page, it needs a patched NOVA. As it wasn't possible to add the additional fields at the end of the hip, other configuration also might need the new NOVA.
The patched NOVA can be found at https://github.com/jklmnn/NOVA/commit/07da624bf12e964389ed4bf6d1a507272876bc46 or when https://github.com/alex-ab/NOVA/pull/2 gets merged.

Furthermore it requires the changes proposed in #2242 (https://github.com/alex-ab/genode/tree/issue_2242). Since the Grub2 used here is insufficient to provide a framebuffer, you can cherry pick https://github.com/jklmnn/genode/commit/172e084f7f1fa5e2e5f9e86a62f8c38909e4993f that adds a grub with the `efi_uga` and `efi_gop` modules. It also increases the efi image size to fit the new grub.

### DIFF
--- a/repos/base-nova/include/nova/syscall-generic.h
+++ b/repos/base-nova/include/nova/syscall-generic.h
@@ -123,6 +123,16 @@ namespace Nova {
 		uint32_t const tsc_freq;    /* time-stamp counter frequency in kHz     */
 		uint32_t const bus_freq;    /* bus frequency in kHz                    */
 
+                struct fb_desc_t {
+                    uint64_t addr;
+                    uint32_t pitch;
+                    uint32_t width;
+                    uint32_t height;
+                    uint8_t bpp;
+                    uint8_t type;
+                    uint8_t reserved[2];
+                } __attribute__((packed)) fb_desc;
+
 		bool has_feature_vmx() const { return feature_flags & (1 << 1); }
 		bool has_feature_svm() const { return feature_flags & (1 << 2); }
 
--- a/repos/os/run/multiboot2_fb.run
+++ b/repos/os/run/multiboot2_fb.run
@@ -1,0 +1,277 @@
+#
+# Build
+#
+
+if {[have_spec odroid_xu]} {
+	puts "Run script does not support this platform."
+	exit 0
+}
+
+set build_components {
+	core init
+	drivers/timer
+	server/nitpicker app/pointer app/status_bar app/xray_trigger
+	server/liquid_framebuffer app/launchpad app/scout
+	test/nitpicker server/nitlog
+	drivers/input/dummy
+        drivers/framebuffer/spec/multiboot2
+	server/report_rom server/rom_filter
+}
+
+proc gpio_drv { } { if {[have_spec rpi] && [have_spec hw]}  { return hw_gpio_drv }
+                    if {[have_spec rpi] && [have_spec foc]} { return foc_gpio_drv }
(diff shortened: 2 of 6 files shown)


alex-ab:
Great ! I will have a look and test it as soon as I find the time. From my point of view, we will need to change a bit your implementation, to let it fit better with other x86 kernels. Using the hip of nova directly in the framebuffer renders the driver useless for them. On the #2242 branch I already introduced a specific rom in core, which contains the acpi rsdp pointer. (So reading the information from the Nova HIP and adding it the kernel independent ROM). All information you need for the efi framebuffer should be added there. This ROM would be filled with the required information by core in a kernel specific way (core is anyway kernel specific) but the final ROM would be kernel-independent for the framebuffer driver. Maybe you can rename the driver to uefi_fb_drv or so. I open for other opinions.

Again, great job !

jklmnn:
Well, first about the naming. The according to the grub2 source it supports multiple drivers to provide a framebuffer which do not depend on (u)efi. We called it `efi_fb` before but since we use an explicit multiboot2 feature that transparently provides a framebuffer and hides the underlying driver we decided to call it as it is called now.
From `/include/grub/video.h`:
```
typedef enum grub_video_driver_id
{
      GRUB_VIDEO_DRIVER_NONE,
      GRUB_VIDEO_DRIVER_VBE,
      GRUB_VIDEO_DRIVER_EFI_UGA,
      GRUB_VIDEO_DRIVER_EFI_GOP,
      GRUB_VIDEO_DRIVER_SM712,
      GRUB_VIDEO_DRIVER_VGA,
      GRUB_VIDEO_DRIVER_CIRRUS,
      GRUB_VIDEO_DRIVER_BOCHS,
      GRUB_VIDEO_DRIVER_SDL,
      GRUB_VIDEO_DRIVER_SIS315PRO,
      GRUB_VIDEO_DRIVER_RADEON_FULOONG2E,
      GRUB_VIDEO_DRIVER_COREBOOT,
      GRUB_VIDEO_DRIVER_IEEE1275,
      GRUB_VIDEO_ADAPTER_CAPTURE,
      GRUB_VIDEO_DRIVER_XEN,
      GRUB_VIDEO_DRIVER_RADEON_YEELOONG3A
} grub_video_driver_id_t;
```

About the hip, I'm looking into the core rom, adding something kernel specific to something already kernel specific makes sense (unless we add this feature to other kernels).

One question I have is, is the location in the Genode source tree okay? I wasn't sure about that.

jklmnn:
I made the proposed changes in https://github.com/jklmnn/genode/commit/5ce20cf4cfae1492ce15e94e7588905fd38e6eff. This however doesn't merge conflict free as it is build upon https://github.com/alex-ab/genode/commit/78056704c37f495b782a77294dc14540b6bc371b. The adapted NOVA is at https://github.com/jklmnn/NOVA/commit/df92d9f8fe9297c2919fd195ff444b3b8bdb6e89.
I mangled the framebuffer specs into the `size` and `aux` fields of `mem_desc` as I didn't see a better way.

alex-ab:
Incorporated the commits to #2242 issue.

PR #2430, opened 2017-05-26
Update Muen

Kensan:
Use latest Muen upstream version and adapt hw_muen VirtualBox to reflect changes to subject state. Also update hw_x86_64_muen documentation/tutorial and adjust RUN_OPTs.

Note: Upstream changes wrt. hardware specification require minor adaptations to the @genodelabs autopilot buildhost. Generated hardware XML specifications have been moved to **policy/hardware** and each hardware target must now provide a corresponding platform specification in **policy/platform**. The purpose of the platform spec is to have uniform naming of serial devices and declare if xHCI debug is supported. See the [platform specification of T430s](https://git.codelabs.ch/?p=muen.git;a=blob;f=policy/platform/lenovo-t430s.xml;h=aace375) for an example.
Also note that the path of the **--image-muen-hardware** RUN_OPT has changed.


chelmuth:
Part of Genode 17.05.

PR #2461, opened 2017-07-04
Do not NULL-terminate readlink result

senier:
The readlink() implementation of Genodes VFS returns the link target as NULL-terminated string. Nothing in POSIX excludes this behavior, but the informative section on readlink in POSIX.1-2008 explicitly mentions that the NULL-termination can be omited:

> Conforming applications should not assume that the returned contents of the symbolic link are null-terminated.

Most implementations seem to actually omit it:

> The readlink() system call does not append a NUL character to	buf. (FreeBSD 11)
> Readlink does not append a NUL character to buf. (Darwin)
> readlink() does not append a null byte to buf. (Linux)

As the Android test cases I'm examining in fact test for the *absence* of a trailing NULL character, there may be well cases where this behavior is assumed. Hence, I suggest to drop the trailing NULL in Genodes VFS implementation for better compatibility.

Note, that I deliberately left the NULL character in the result, but just decremented the returned result length to mitigate unsafe uses of strlen() and friends.

### DIFF
--- a/repos/os/src/lib/vfs/fs_file_system.h
+++ b/repos/os/src/lib/vfs/fs_file_system.h
@@ -477,7 +477,7 @@ class Vfs::Fs_file_system : public File_system
 				    _fs.symlink(dir_handle, symlink_name.base() + 1, false);
 				Fs_handle_guard symlink_guard(*this, _fs, symlink_handle, _handle_space);
 
-				out_len = _read(symlink_guard, buf, buf_size, 0);
+				out_len = _read(symlink_guard, buf, buf_size, 0) - 1;
 
 				return READLINK_OK;
 			}


chelmuth:
I merged your commit to staging but are wondering how the other VFS plugins (e.g., tar) handle `readlink()`. Did you look into these? It may be worthwhile to generate fixup commits if required.

senier:
No, I didn't look into other plugins.

ehmry:
This patch is incompatible with the VFS server which sets the length returned by the VFS library``readlink`` in the ``File_system`` packet length: 
https://github.com/genodelabs/genode/blob/master/repos/os/src/server/vfs/node.h#L126.

I believe the root of the problem here is that the ``ram_fs`` server does not check the length of symlink writes for which I have a fix at 6369399eba74655d8db11152c00ceb939bbcd549.

@senier the ``ram_fs`` and ``vfs`` servers are almost redundant but its is useful to have the both to test the VFS library against. When working with the VFS library it is important to consider a scenario where a ``File_system`` server may be proxied by the ``vfs`` server, which in the case of you patch would reduce the length of the symlink by a byte for every ``vfs`` server the readlink traverses. The ``vfs`` server may seem to muddy the water at first but it actually helps discipline the library.

senier:
Thanks for the explanation. I tested your solution in https://github.com/genodelabs/genode/commit/6369399eba74655d8db11152c00ceb939bbcd549 and it indeed fixes #2461.

chelmuth:
Fixed in master.

PR #2464, opened 2017-07-20
base: added vsnprintf.

jklmnn:
Added vsnprintf to use va_lists directly with snprintf.

### DIFF
--- a/repos/base/include/base/snprintf.h
+++ b/repos/base/include/base/snprintf.h
@@ -32,6 +32,8 @@ namespace Genode {
 	 */
 	inline int snprintf(char *dst, size_t dst_size, const char *format, ...)
 	           __attribute__((format(printf, 3, 4)));
+	inline int vsnprintf(char *dst, size_t dst_size, const char *format, va_list list)
+	           __attribute__((format(printf, 3, 0)));
 }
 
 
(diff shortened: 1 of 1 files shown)


nfeske:
One year ago, we decided to abolish format strings from Genode's base API. (see http://genode.org/documentation/release-notes/16.05#New_log-output_facilities) We still keep the functions around as some portions of the code still rely on them but they will ultimately be removed. For this reason, we are not going to extend them.

If you need `vsnprintf` in your application, please use the libc, which implements proper format string support (as opposed to the rather minimalistic subset provided by Genode's version). Should your component be a genuine Genode component, for which you'd like to avoid the dependency from the libc, please consider using the `Genode::String` constructor for assembling character strings.

jklmnn:
As I'm creating a dde_linux driver, I think I'll go with the libc version, thanks.

nfeske:
For a dde_linux driver, you may better use the printf implementation locally provided within the dde_linux repository. See `src/lx_kit/printf.cc` and `src/include/lx_emul/printf.h`. Pull requests for this version are welcome because it is meant to stay. ;-)

ssumpf:
@jklmnn: For dde_linux it is recommended to implement the functions directly in dde_linux. `vsnprintf' is already implemented by the USB driver (src/lib/usb/lx_emul.cc) and the WIFI driver (src/lib/wifi/lxcc_emul.cc), you could simply copy a version into your driver, or since the function is used by more than one driver, we could also move into the lx_kit directory. Please don not link your driver against libc.

PR #2479, opened 2017-08-12
base-linux: Log error message for too long files

senier:
The filename buffer of a dataspace in base-linux is limited to
40 bytes. When using file names longer than this, the remainder
gets dropped silently. Add an error message to aid debugging this
case.


nfeske:
Thank you. Merged to staging - with the minor adjustment of starting the message lower-case to be consistent with the existing practice.

chelmuth:
Merged as ee40d29b914020c6cfb048892453378d17d0f22a to master.

PR #2480, opened 2017-08-12
global.mk: Make build messages configurable

senier:
This is helpful for disabling messages in etc/tools.conf by
setting it to e.g.

   MSG_LINK = @true ""

This results in much shorter and less cluttered logs in automatic
builds.


chelmuth:
Merged as 23db75deff662130bdbb97c949f3c14862295c03 to master.

PR #2524, opened 2017-09-20
test/libc_ffat: Test ftruncate on unlinked file

senier:
This amends the libc_vfs test with a case that crashes ram_fs in my setup:

```
[init -> test-libc_vfs] calling ftruncate(fd, 42)
no RM attachment (READ pf_addr=0x0 pf_ip=0x0 from pager_object: pd='init -> ram_fs' thread='ep')
page fault, pd='init -> ram_fs' thread='ep' cpu=0 ip=0x0 address=0x0 stack pointer=0xa04fef18 qualifiers=0x4 irUwp reason=1
Error: Test execution timed out
```
I tested this on x86_64 (base-nova as well as base-linux).

The issue arises when unlinking the file behind an *opened* fd and then performing an ftruncate() on it. The expected behavior is a properly resized file accessible by the fd.

### DIFF
--- a/repos/libports/src/test/libc_vfs/main.cc
+++ b/repos/libports/src/test/libc_vfs/main.cc
@@ -260,7 +260,7 @@ static void test(Genode::Xml_node node)
 		CALL_AND_CHECK(ret, close(fd), ret == 0, "");
 
 		/* test 'O_TRUNC' flag */
-		CALL_AND_CHECK(fd, open(file_name4, O_WRONLY | O_TRUNC), fd >= 0, "file_name=%s", file_name4);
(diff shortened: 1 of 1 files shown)


nfeske:
Thank you for the test case. I suspect that the problem is related to the following comment: https://github.com/genodelabs/genode/blob/master/repos/os/src/server/ram_fs/main.cc#L387

chelmuth:
Is this still valid?

cproc:
@chelmuth: the ram_fs crash is fixed in issue #2536, but the test still fails, because reference counting is not implemented.

chelmuth:
According to the release notes of 18.08, the VFS is now on par with ram_fs. Does the test also fail when using the VFS server instead?

nfeske:
This issue is not a pressing problem and has become stale.

PR #2547, opened 2017-10-25
RFC platform_drv/x86: Switch to ECAM/MMCONF

reet-:
This PR is intended as RFC.

Factor out system reset functionality to separate file. This code requires port I/O to operate.

Switch the x86 PCI config space access method from Configuration Access Mechanism (CAM, port I/O) to Enhanced Configuration Access Mechanism (ECAM, mmio). The base address of the PCI configuration space is acquired by mapping the ACPI ROM and reading the bdf node.

The code is implemented in a way to make it work with Muen subject monitor (SM) device emulation and also general x86 targets.

### DIFF
--- a/repos/os/src/drivers/platform/spec/x86/pci_config_access.h
+++ b/repos/os/src/drivers/platform/spec/x86/pci_config_access.h
@@ -1,6 +1,7 @@
 /*
  * \brief  Interface for accessing PCI configuration registers
  * \author Norman Feske
+ * \author Reto Buerki
  * \date   2008-01-29
  */
 
@@ -15,8 +16,11 @@
 #define _X86_PCI_CONFIG_ACCESS_H_
 
 #include <util/bit_array.h>
-#include <io_port_session/connection.h>
+#include <base/attached_io_mem_dataspace.h>
+#include <base/attached_rom_dataspace.h>
 #include <platform_device/platform_device.h>
+
+using namespace Genode;
 
 namespace Platform {
 
@@ -24,51 +28,23 @@ namespace Platform {
 	{
 		private:
 
-			Genode::Env &_env;
-
-			enum { REG_ADDR = 0xcf8, REG_DATA = 0xcfc, REG_SIZE = 4 };
+			Attached_io_mem_dataspace &_pciconf;
 
 			/**
-			 * Request interface to access an I/O port
-			 */
-			template <unsigned port>
-			Genode::Io_port_session *_io_port()
-			{
-				/*
-				 * Open I/O-port session when first called.
-				 * The number of instances of the io_port_session_client
-				 * variable depends on the number of instances of the
-				 * template function.
-				 *
-				 * Thanks to this mechanism, the sessions to the PCI
-				 * ports are created lazily when PCI service is first
-				 * used. If the PCI bus driver is just started but not
-				 * used yet, other processes are able to access the PCI
-				 * config space.
-				 *
-				 * Once created, each I/O-port session persists until
-				 * the PCI driver gets killed by its parent.
-				 */
-				static Genode::Io_port_connection io_port(_env, port, REG_SIZE);
-				return &io_port;
-			}
-
-			/**
-			 * Generate configuration address
+			 * Calculate device offset from BDF
 			 *
 			 * \param bus       target PCI bus ID  (0..255)
 			 * \param device    target device ID   (0..31)
 			 * \param function  target function ID (0..7)
-			 * \param addr      target byte within targeted PCI config space (0..255)
 			 *
-			 * \return configuration address (written to REG_ADDR register)
+			 * \return device base address
 			 */
-			unsigned _cfg_addr(int bus, int device, int function, int addr) {
-				return ( (1        << 31) |
-				         (bus      << 16) |
-				         (device   << 11) |
-				         (function <<  8) |
-				         (addr      & ~3) ); }
+			unsigned _dev_base(int bus, int device, int function)
+			{
+				return ((bus      << 20) |
+				        (device   << 15) |
+				        (function << 12));
+			}
 
 			Genode::Bit_array<256> _used { };
 
@@ -81,7 +57,10 @@ namespace Platform {
 
 		public:
 
-			Config_access(Genode::Env &env) : _env(env) { }
+			Config_access(Attached_io_mem_dataspace &pciconf)
+			: _pciconf(pciconf) { }
+
+			Config_access(Config_access &c) : _pciconf(c._pciconf) { }
 
 			/**
 			 * Read value from config space of specified device/function
@@ -100,26 +79,37 @@ namespace Platform {
 			              unsigned char addr, Device::Access_size size,
 			              bool track = true)
 			{
-				/* write target address */
-				_io_port<REG_ADDR>()->outl(REG_ADDR, _cfg_addr(bus, device, function, addr));
+				unsigned ret;
+				char const * const field = _pciconf.local_addr<char>() + _dev_base(bus, device, function) + addr;
 
-				/* return read value */
+				/*
+				 * Memory access code is implemented in a way to make it work
+				 * with Muen subject monitor (SM) device emulation and also
+				 * general x86 targets. On Muen, the simplified device
+				 * emulation code (which also works for Linux) always returns
+				 * 0xffff in EAX to indicate a non-existing device. Therefore,
+				 * we enforce the usage of EAX in the following assembly
+				 * templates. Also clear excess bits before return to guarantee
+				 * the requested size.
+				 */
 				switch (size) {
 				case Device::ACCESS_8BIT:
 					if (track)
 						_use_register(addr, 1);
-
-					return _io_port<REG_DATA>()->inb(REG_DATA + (addr & 3));
+					asm volatile("movb %1,%%al" :"=a" (ret) :"m" (*((volatile unsigned char *)field)) :"memory");
+					return ret & 0xff;
 				case Device::ACCESS_16BIT:
 					if (track)
 						_use_register(addr, 2);
 
-					return _io_port<REG_DATA>()->inw(REG_DATA + (addr & 2));
+					asm volatile("movw %1,%%ax" :"=a" (ret) :"m" (*(volatile unsigned short *)field) :"memory");
+					return ret & 0xffff;
 				case Device::ACCESS_32BIT:
 					if (track)
 						_use_register(addr, 4);
 
-					return _io_port<REG_DATA>()->inl(REG_DATA);
+					asm volatile("movl %1,%%eax" :"=a" (ret) :"m" (*(volatile unsigned int *)field) :"memory");
+					return ret;
 				default:
 					return ~0U;
 				}
@@ -141,29 +131,30 @@ namespace Platform {
 			           unsigned value, Device::Access_size size,
 			           bool track = true)
 			{
-				/* write target address */
-				_io_port<REG_ADDR>()->outl(REG_ADDR, _cfg_addr(bus, device,
-				                                               function, addr));
+				char const * const field = _pciconf.local_addr<char>() + _dev_base(bus, device, function) + addr;
 
-				/* write value to targeted address */
+				/*
+				 * Write value to targeted address, see read() comment above
+				 * for an explanation of the assembly templates
+				 */
 				switch (size) {
 				case Device::ACCESS_8BIT:
 					if (track)
 						_use_register(addr, 1);
 
-					_io_port<REG_DATA>()->outb(REG_DATA + (addr & 3), value);
+					asm volatile("movb %%al,%1" : :"a" (value), "m" (*(volatile unsigned char *)field) :"memory");
 					break;
 				case Device::ACCESS_16BIT:
 					if (track)
 						_use_register(addr, 2);
 
-					_io_port<REG_DATA>()->outw(REG_DATA + (addr & 2), value);
+					asm volatile("movw %%ax,%1" : :"a" (value), "m" (*(volatile unsigned char *)field) :"memory");
 					break;
 				case Device::ACCESS_32BIT:
 					if (track)
 						_use_register(addr, 4);
 
-					_io_port<REG_DATA>()->outl(REG_DATA, value);
+					asm volatile("movl %%eax,%1" : :"a" (value), "m" (*(volatile unsigned char *)field) :"memory");
 					break;
 				}
 			}
@@ -180,32 +171,6 @@ namespace Platform {
 				default:
 					return true;
 				}
-			}
-
-			bool reset_support(unsigned reg, unsigned reg_size) const
-			{
-				return (REG_ADDR <= reg) &&
-				       reg + reg_size <= REG_ADDR + REG_SIZE;
-			}
-
-			bool system_reset(unsigned reg, unsigned long long value,
-			                  const Device::Access_size &access_size)
-			{
-				switch (access_size) {
-				case Device::ACCESS_8BIT:
-					_io_port<REG_ADDR>()->outb(reg, value);
-					break;
-				case Device::ACCESS_16BIT:
-					_io_port<REG_ADDR>()->outw(reg, value);
-					break;
-				case Device::ACCESS_32BIT:
-					_io_port<REG_ADDR>()->outl(reg, value);
-					break;
-				default:
-					return false;
-				}
-
-				return true;
 			}
 	};
 }
--- a/repos/base-nova/src/core/pd_session_support.cc
+++ b/repos/base-nova/src/core/pd_session_support.cc
@@ -45,8 +45,10 @@ void Pd_session_component::map(addr_t virt, addr_t size)
 	                   addr_t                region_offset) -> addr_t
 	{
 		Dataspace_component * dsc = region ? region->dataspace() : nullptr;
-		if (!dsc)
-			return ~0UL;
+		if (!dsc) {
+			struct No_dataspace{};
+			throw No_dataspace();
+		}
 
 		Mapping mapping = Region_map_component::create_map_item(region_map,
 		                                                        region,
@@ -87,9 +89,13 @@ void Pd_session_component::map(addr_t virt, addr_t size)
 		return mapped;
 	};
 
-	while (size) {
-		addr_t mapped = _address_space.apply_to_dataspace(virt, lambda);
-		virt         += mapped;
-		size          = size < mapped ? size : size - mapped;
+	try {
+		while (size) {
+			addr_t mapped = _address_space.apply_to_dataspace(virt, lambda);
+			virt         += mapped;
+			size          = size < mapped ? size : size - mapped;
+		}
+	} catch (...) {
+		error(__func__, " failed ", Hex(virt), "+", Hex(size));
 	}
 }
--- a/repos/base-okl4/src/core/platform.cc
+++ b/repos/base-okl4/src/core/platform.cc
@@ -164,7 +164,7 @@ Platform::Platform() :
 	                                           stack_area_virtual_size());
 
 	_vm_start = 0x1000;
-	_vm_size  = 0xb0000000 - 0x1000;
+	_vm_size  = 0xc0000000 - _vm_start;
 
 	log(_rom_fs);
 
--- a/repos/base-pistachio/src/core/platform.cc
+++ b/repos/base-pistachio/src/core/platform.cc
@@ -550,13 +550,11 @@ void Platform::_setup_basics()
 
 	/* remove KIP area from region and IO_MEM allocator */
 	remove_region(Region((addr_t)kip, (addr_t)kip + kip_size), _region_alloc);
-	remove_region(Region((addr_t)kip, (addr_t)kip + kip_size), _io_mem_alloc);
 
 	/* remove utcb area */
 	addr_t utcb_ptr = (addr_t)Platform_pd::_core_utcb_ptr;
 
 	remove_region(Region(utcb_ptr, utcb_ptr + L4_UtcbAreaSize (kip)), _region_alloc);
-	remove_region(Region(utcb_ptr, utcb_ptr + L4_UtcbAreaSize (kip)), _io_mem_alloc);
 
 	/* remove core program image memory from region allocator */
 	addr_t img_start = (addr_t) &_prog_img_beg;
--- a/repos/base/run/platform_drv.inc
+++ b/repos/base/run/platform_drv.inc
@@ -68,9 +68,14 @@ proc append_platform_drv_build_components {} {
 proc platform_drv_boot_modules {} {
 	set drv_boot_modules ""
 	lappend_if [have_platform_drv] drv_boot_modules platform_drv
-	lappend_if [have_spec x86]     drv_boot_modules acpi_drv
 	lappend_if [have_spec x86]     drv_boot_modules report_rom
-	lappend_if [have_spec muen]    drv_boot_modules acpi
+
+	if {[have_spec muen]} {
+		lappend drv_boot_modules acpi
+	} else {
+		lappend drv_boot_modules acpi_drv
+	}
+
 	return $drv_boot_modules
 }
 
@@ -114,20 +119,15 @@ proc platform_drv_add_routing {} {
 
 
 proc platform_drv_config_config {} {
-	if {[have_spec acpi] || [have_spec arm] || [have_spec muen]} {
-		return {
-		<config>}
-	}
-
 	return {
-		<config acpi="no">}
+		<config>}
 }
 
 
 proc platform_drv_config {} {
 	set drv_config ""
 
-	if {[have_spec x86]} {
+	if {[have_spec x86] && ![have_spec muen]} {
 
 		append drv_config {
 	<start name="acpi_drv" } [platform_drv_priority] { caps="400" >
--- a/repos/base/run/platform_drv.inc
+++ b/repos/base/run/platform_drv.inc
@@ -54,8 +54,8 @@ proc audio_drv_binary { } {
 proc platform_drv_build_components {} {
 	set drv_build_components ""
 	lappend_if [have_platform_drv] drv_build_components drivers/platform
-	lappend_if [have_spec acpi]    drv_build_components drivers/acpi
-	lappend_if [have_spec acpi]    drv_build_components server/report_rom
+	lappend_if [have_spec x86]     drv_build_components drivers/acpi
+	lappend_if [have_spec x86]     drv_build_components server/report_rom
 	return $drv_build_components
 }
 
@@ -68,8 +68,8 @@ proc append_platform_drv_build_components {} {
 proc platform_drv_boot_modules {} {
 	set drv_boot_modules ""
 	lappend_if [have_platform_drv] drv_boot_modules platform_drv
-	lappend_if [have_spec acpi]    drv_boot_modules acpi_drv
-	lappend_if [have_spec acpi]    drv_boot_modules report_rom
+	lappend_if [have_spec x86]     drv_boot_modules acpi_drv
+	lappend_if [have_spec x86]     drv_boot_modules report_rom
 	lappend_if [have_spec muen]    drv_boot_modules acpi
 	return $drv_boot_modules
 }
@@ -104,7 +104,7 @@ proc platform_drv_policy {} {
 
 proc platform_drv_priority {} { return "" }
 proc platform_drv_add_routing {} {
-	if {[have_spec acpi]} {
+	if {[have_spec x86]} {
 		return {
 			<service name="ROM" label="system"> <child name="acpi_report_rom"/> </service>}
 	}
@@ -127,7 +127,7 @@ proc platform_drv_config_config {} {
 proc platform_drv_config {} {
 	set drv_config ""
 
-	if {[have_spec acpi]} {
+	if {[have_spec x86]} {
 
 		append drv_config {
 	<start name="acpi_drv" } [platform_drv_priority] { caps="400" >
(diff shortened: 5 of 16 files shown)


repos/os/src/drivers/platform/spec/x86/pci_config_access.h
```diff
@@ -81,7 +60,17 @@ namespace Platform {
...
 
-			Config_access(Genode::Env &env) : _env(env) { }
+			Config_access(Env &env) : _env(env) {
+				Attached_rom_dataspace info(_env, "acpi");
+				Xml_node xml(info.local_addr<char>(), info.size());
+				for (unsigned i = 0; i < xml.num_sub_nodes(); i++) {
+					Xml_node node = xml.sub_node(i);
+
+					if (node.has_type("bdf")) {
+						node.attribute("base").value(&_base);
```
chelmuth: `Xml_node` methods may throw, for example, `Nonexistent_attribute`, which should be catched. For convenience, you may use `Xml_node::attribute_value`. There's also `Xml::for_each_sub_node()` which seems to fit your use case.

repos/os/src/drivers/platform/spec/x86/pci_config_access.h
```diff
@@ -100,26 +89,34 @@ namespace Platform {
 			              unsigned char addr, Device::Access_size size,
 			              bool track = true)
 			{
-				/* write target address */
-				_io_port<REG_ADDR>()->outl(REG_ADDR, _cfg_addr(bus, device, function, addr));
+				unsigned ret;
+				const addr_t mmconf = _base + _dev_base(bus, device, function);
```
chelmuth: We prefer to post-fix cv-qualifiers (const, volatile) to the type, so this line should start with `addr_t const mmconf...`

repos/os/src/drivers/platform/spec/x86/pci_config_access.h
nfeske: Let me add that the `Attached_rom_dataspace` has an `xml()` method, which alleviates the need to manually construct a `Xml_node`.

repos/os/src/drivers/platform/spec/x86/pci_config_access.h
```diff
@@ -100,26 +89,34 @@ namespace Platform {
...
+				 * general x86. Furthermore, clear excess bits before return to
+				 * guarantee requested size.
+				 */
 				switch (size) {
 				case Device::ACCESS_8BIT:
 					if (track)
 						_use_register(addr, 1);
-
-					return _io_port<REG_DATA>()->inb(REG_DATA + (addr & 3));
+					asm volatile("movb %1,%%al" :"=a" (ret) :"m" (*((volatile unsigned char *)field)) :"memory");
```
chelmuth: IMO dereferencing a `unsigned char *` pointer should suffice here. Do you agree?

repos/os/src/drivers/platform/spec/x86/pci_config_access.h
reet-: Thanks for the input, I will simplify the code accordingly.

repos/os/src/drivers/platform/spec/x86/pci_config_access.h
reet-: Will do.

repos/os/src/drivers/platform/spec/x86/pci_config_access.h
reet-: Instead of `volatile unsigned char *`?

Will drop superfluous parenthesis in the `ACCESS_8BIT` case.

repos/os/src/drivers/platform/spec/x86/pci_config_access.h
```diff
@@ -141,29 +138,29 @@ namespace Platform {
 			           unsigned value, Device::Access_size size,
 			           bool track = true)
 			{
-				/* write target address */
-				_io_port<REG_ADDR>()->outl(REG_ADDR, _cfg_addr(bus, device,
-				                                               function, addr));
+				const addr_t mmconf = _base + _dev_base(bus, device, function);
+				Attached_io_mem_dataspace mmio(_env, mmconf, MMCONF_REGION_SIZE);
```
alex-ab: Please open this Attached_io_mem dataspace once in the constructor and don't create and destruct this io_mem session with every read/write.

reet-: This is a per-device dataspace. Can you please elaborate?

alex-ab: Mmh, unfortunately this class is not per device instead for all pci devices. Before your proposal change this made sense, since there was only one specific io port range for all pci devices. Now it changed.

This "_base + _dev_base()" construct assumes, that there is only one contiguous memory range for a given mmconf range - so it would make sense to open this dataspace once and attach all once.

repos/os/src/drivers/platform/spec/x86/pci_config_access.h
chelmuth: Sorry, my note was somewhat misleading. What I actually question is the use of inline assembler for a simple `unsigned char ret = *(unsigned char volatile *)field`. Is there any need for the assembler code?

repos/os/src/drivers/platform/spec/x86/pci_config_access.h
alex-ab: I think this code is not correct. The "bdf" element may appear more than once. This "bdf" node correlates to the mcfg table, which may report several mmconf regions for some part of the pci bdf space. So, you will have to add here a Object which stores this base address and the size, and the pci range this mmconf is fore.

alex-ab: I meant a list of such mmconf object ranges and you would have to adjust this "_base + _dev_base(bdf)" calculation by going effectively through this list.

repos/os/src/drivers/platform/spec/x86/pci_config_access.h
reet-: You are totally right, for the non-Muen case a dereference would probably suffice. On Muen however, our simplified device emulation code always returns `0xffff` in RAX to indicate a non-existing device (see https://git.codelabs.ch/?p=muen.git;a=blob;f=components/sm/src/devices-pciconf.adb;h=a953b38c957214a2865178b759b9e14c53687a8b;hb=dd59337163c174d584b5e1245b9278e2a109e5f8#l662). 

This simple method works well with Linux guests. 

Also, the inline assembly is inspired by the `read*` functions of the Linux kernel (`arch/x86/include/asm/io.h`) which also uses assembly on x86.

repos/os/src/drivers/platform/spec/x86/pci_config_access.h
reet-: As discussed on IRC, I will adjust the code as follows:
- Move handling of PCI config space mapping to early platform drv boot code (`Platform::Root)`
  - Throw error if multiple bdf nodes exist
  - This is a simplification since we have not yet encountered systems with multiple PCI segment groups
  - Throw error if no node for segment 0:0.0 exists
  - Map complete PCI segment group region for 0:0.0 early in the platform drv boot code
- Pass on PCI config space mapping reference to session, device, config access

chelmuth:
@alex-ab @reet- what's the state of this patches? I don't object to switch MMCONF but we have to sort out the issues with Qemu *pc* first, i.e., switch to *q35*. IMO keeping both configuration mechanisms is no option though.

reet-:
@chelmuth I was working on other things the last few days. Hoping to get back to it this week. I'll keep you posted.

alex-ab:
I adjusted Qemu to use q35 (for MMCONF/ECAM support). Additionally I adjusted the acpi_drv/platform_drv interplay to provide now also for non ACPI kernels (fiasco, pistachio, okl4) the MMCONF/ECAM region, so that the platform_drv can operate properly. On Qemu it looks ok (tested nic and usb driver), on real hardware however it is not going to work. The ECAM regions are on the machine I had in a range between 3G - 4G, which some of the kernels and/or core don't give access to. Further investigation in this direction would be required to resolve this issue.

chelmuth:
If I understand your patch for pistachio correctly, the kernel puts the KIP in the ECAM region and I doubt we are going to fix this in the contrib kernel. Does fiasco also have comparable issues?

alex-ab:
The patch was required on Qemu for Pistachio:

```
[init -> platform_drv] ECAM/MMCONF range 0:0.0-ff:1f.7 - addr [b0000000,c0000000)
```

Core on pistachio kernel:

```
Error: kip [bff00000,bff00120)
Error: utcb area [bf000000,bf001000)
```

Core on Fiasco seems fine:
```
Error: vm area (for 1:1 mappings) [00001000,c0000000)
```

On real hardware all 3 fail to provide access to the region (3G-4G physical). I presume using MMCONF/ECAM for our oldies is not a good idea. We would have to make sure the kernel can handle/know these regions, adjust all 3 Genode Core's, maintain the contrib kernel code and so on

chelmuth:
Sorry, but I still did not get what issue you refer to with

> On real hardware all 3 fail to provide access to the region (3G-4G physical).

I ran three tests with current staging and the following patch on our 64-bit test machine. All kernels handle the MMIO region for the network adapter correctly. Please correct me if I'm wrong. (The log files are attached.)
[fiasco.log](https://github.com/genodelabs/genode/files/1492161/fiasco.log)
[macho-fiasco.log](https://github.com/genodelabs/genode/files/1492162/macho-fiasco.log)
[macho-okl4.log](https://github.com/genodelabs/genode/files/1492163/macho-okl4.log)



```diff
diff --git a/repos/base/src/core/io_mem_session_component.cc b/repos/base/src/core/io_mem_session_component.cc
--- a/repos/base/src/core/io_mem_session_component.cc
+++ b/repos/base/src/core/io_mem_session_component.cc
@@ -64,6 +64,12 @@ Io_mem_session_component::_prepare_io_mem(const char      *args,
 	/* request local mapping */
 	addr_t local_addr = _map_local(base, size);
 
+Genode::log("\033[00;37m",
+            "IOMEM: ", Genode::Hex(base), "-", Genode::Hex(base+(size-1)),
+            " @ ", Genode::Hex(local_addr),
+            " for \"", label_from_args(args), "\"",
+            "\033[0m");
+
 	return Dataspace_attr(size, local_addr, base, _cacheable, req_base);
 }
 
diff --git a/repos/os/src/drivers/platform/spec/x86/pci_device_component.h b/repos/os/src/drivers/platform/spec/x86/pci_device_component.h
--- a/repos/os/src/drivers/platform/spec/x86/pci_device_component.h
+++ b/repos/os/src/drivers/platform/spec/x86/pci_device_component.h
@@ -52,7 +52,12 @@ class Platform::Device_component : public Genode::Rpc_object<Platform::Device>,
 			public:
 				Io_mem (Genode::Env &env, Genode::addr_t base,
 			            Genode::size_t size, bool wc)
-				: Genode::Io_mem_connection(env, base, size, wc) { }
+				: Genode::Io_mem_connection(env, base, size, wc)
+				{
+					Genode::log("\033[00;37m",
+					            "IOMEM: ", Genode::Hex(base), "-", Genode::Hex(base+(size-1)),
+					            "\033[0m");
+				}
 		};
 
 		enum {
```

alex-ab:
Yeah, current staging is fine. Just try the issue branch.

alex-ab:
If you would request for all possible PCI devices the I/O mem you would also fail, that is what the issue branch does - getting all I/O mem for the PCI devices.

reet-:
This sounds to me that just dropping port I/O access is not an option, but instead the mechanism should be made configurable. On the other hand, I don't think this is easily possible with the current code.

How should we proceed?

chelmuth:
Still, I'm not fond of adding MMCONF as a configurable alternative. IMO we should walk into this direction and clear obstacles on the way together. @alex-ab do you have time on Thursday to quickly present me any issues with OKL4 and Fiasco? Also, I may come up with an idea to "repair" the Pistachio kernel protocol for KIP and root-task UTCB. If not, I vote for removing this platform. With regard to OKL4 and Fiasco, I think different because of the leanness of OKL4 and the debugging facilities of Fiasco.

alex-ab:
As said, we need to investigate and need time. My last finding for today, for okl4 it seems that the scanning behind a bridge fails with a pagefault in the platform_drv.

alex-ab:
The whole ECAM memory range on one of our test machines is not reported completely by the acpi tables (verified also by a linux boot, which has the same problem). That means we have to be a more robust in the platform driver to handle also such machines with incomplete/broken acpi tables. Namely we will have to replace the simple attached_io_mem for the pci config space in the platform driver with another class, which knows about the missing parts and on can handle it.

reet-:
@alex-ab Thanks for looking into this. Can you please give an example of what is reported on this platform and what the correct values are?

alex-ab:
Based on my issue branch and running pci.run on the target machine (okl4 - kernel is irrelevant actually, apci driver report is same and is correct according to linux boot, which also complains):

```
[init -> acpi_drv] Found MCFG
[init -> acpi_drv] MCFG BASE 0xe0000000 seg 0x0 bus 0x0-0x10
[init -> acpi_drv] Found MADT
[init -> acpi_drv] MADT IRQ 0 -> GSI 2 flags: 5
[init -> acpi_drv] MADT IRQ 9 -> GSI 9 flags: 13
[init -> platform_drv] ECAM/MMCONF range 0:0.0-10:1f.7 - phys_addr [e0000000,e1100000)
[init -> platform_drv] ECAM/MMCONF range 0:0.0-10:1f.7 - phys_addr [e0000000,e1100000) attached to [0b000000,0c100000)
[init -> platform_drv] Warning: MADT IRQ 9-> GSI 9 flags 13 ignored
[init -> platform_drv] Warning: MADT IRQ 0-> GSI 2 flags 5 ignored
[init -> platform_drv] Warning: 0:1f.0 ignore ACPI IRQ routing: 0->19
...
[init -> platform_drv] Warning: 0:1.0 ignore ACPI IRQ routing: 255->16
[init -> platform_drv] iterate now devices behind bridge 0:1c.0
[init -> platform_drv] device on pci bus behind bridge 5:0.0
...
[init -> platform_drv] device on pci bus behind bridge 5:1f.7
[init -> platform_drv] iterate now devices behind bridge 0:1c.0 done
[init -> platform_drv] iterate now devices behind bridge 0:1e.0
[init -> platform_drv] device on pci bus behind bridge 11:0.0
no RM attachment (READ pf_addr=0xc100000 pf_ip=0x102d48b from pager_object: pd='init -> platform_drv' thread='ep') 
Warning: invalid signal-context capability
Warning: page-fault, pager_object: pd='init -> platform_drv' thread='ep' ip=0x102d48b pf-addr=0xc100000
Error: Test execution timed out
Makefile:312: recipe for target 'run/pci' failed
```

reet-:
@alex-ab What is the strategy of Linux to correct the invalid ACPI report on this machine? 

The problem should not occur with my initial code, as the MMCONF size is initialized with `0x10000000` (the complete BDF range).

chelmuth:
The Linux kernel fixes the broken table information by assuming the MMCONF region is contiguous beyond the specified limit to cover the bus behind the bridge. But, this is no general assumption and does not expand to  *covering all possible BDFs*. I'm afraid that this assumption would collide with other MMIO resources that may be located in this area of the physical address space on other hardware. Therefore, I suggest to adapt the Linux strategy.

alex-ab:
I cleaned up all commits, incorporated the q35 changes and re-based to current staging [issue_2547_staging](https://github.com/alex-ab/genode/tree/issue_2547_staging). The remaining issue is pistachio, where @chelmuth has some ideas? Otherwise, @reet- please check and @chelmuth consider for review.

nfeske:
@alex-ab What happened to the system reset in your latest commit?

If I understand correctly, with the patch, each device PD gains the authority to assign any PCI device to itself. This is in contrast to the original version where each device PD received this authority on a per-device basis. Therefore, the device PD must now be considered as trusted with regard to cross-domain information flow, which was not the case before. Now, the code of the device PD is simple enough to accept that. But I wonder about possible other implications. E.g., a malicious device driver can issue DMA transactions to the device PD and thereby possibly trick the device PD to map further parts of the PCI config, which would eventually enable the device to gain illegitimate access to another device. Even though this sounds academic, I would feel better if the device PD had no such all-encompassing authority in the first place. If it is not trusted, we don't need to counter concerns like the one above. It may be possible to achieve that in two ways:

* By letting the platform driver directly attach the (portion of the config dataspace) in the device PD's address space.
* By creating a managed dataspace per device PD that has only the device-specific portion of the config space attached. The platform driver would then hand out the managed dataspace to the device PD.

alex-ab:
> @alex-ab What happened to the system reset in your latest commit?

The platform driver don't have anymore I/O ports (for the reset). The logical conclusion is that it can't do the reset anymore.

> If I understand correctly, with the patch, each device PD gains the authority to assign any PCI device to itself.

Not really, no. The device_pd code is part of the platform driver, the platform driver must be trusted.

>  But I wonder about possible other implications. E.g., a malicious device driver can issue DMA transactions to the device PD and thereby possibly trick the device PD to map further parts of the PCI config, which would eventually enable the device to gain illegitimate access to another device. Even though this sounds academic

It is completely bogus. Since you know the code and what the device pd is, I can't imagine how you can conclude such things.

> I would feel better if the device PD had no such all-encompassing authority in the first place. 

We need no device pd in the first place.

nfeske:
> The platform driver don't have anymore I/O ports (for the reset). The logical conclusion is that it can't do the reset anymore.

The commit message does not mention any of that. Neither does the commit updates the `README` in this respect. Also, there is not alternative for implementing this functionality presented.

> Not really, no. The device_pd code is part of the platform driver, the platform driver must be trusted.

The device PD does in fact not need to be trusted. This is by intention. In contrast to the device PD, the platform driver is not targeted by DMA transactions.

> It is completely bogus. Since you know the code and what the device pd is, I can't imagine how you can conclude such things.

Try to put yourself into the shoes of an evaluator of our system:

1. The device PD has authority over the entire PCI config space.
2. The device PD contains executable code.
3. The device PD is targeted by DMA transactions initiated by untrusted (and potentially malicious) device-driver code.

**What could possibly go wrong?**

It does not matter that you think this line of argumentation is "bogus". What matters is that those concerns are easy to raise (by an evaluator) and hard to defend against (by me). Hence, such concerns must be avoided in the first place. I am not nitpicking to provoke you!

alex-ab:
>The device PD has authority over the entire PCI config space.
You are wrong. The platform driver has authority.

> The device PD contains executable code.
You are wrong. The device pd contains noooooooooooooooooooo executable code.

alex-ab:
> The commit message does not mention any of that. Neither does the commit updates the README in this respect. Also, there is not alternative for implementing this functionality presented.

The README got adjusted by 713532205352b131078e767387610dcb6a1db7f6.

alex-ab:
> Also, there is not alternative for implementing this functionality presented.

Acpica does this for you.

chelmuth:
> Also, there is not alternative for implementing this functionality presented.
> 
> Acpica does this for you.

Unfortunately, the removal would degrade the debugging experience with current Sculpt quite a lot as *acpica* had to be started by default. IMO requesting the I/O ports in the reset module of *platform_drv* only does no harm.

chelmuth:
> The remaining issue is pistachio, where @chelmuth has some ideas?

We discussed offline that @alex-ab's patch is correct as the KIP and UTCB regions are virtual and the backing RAM is allocated by the kernel.

nfeske:
> You are wrong. The device pd contains noooooooooooooooooooo executable code.

I missed that you removed the separate device PD component. I would have appreciated a less confronting way of educating me though.

alex-ab:
> Unfortunately, the removal would degrade the debugging experience with current Sculpt quite a lot as acpica had to be started by default.

I can imagine, but degrading debugging experience is IMO no technical argument. The platform driver is dump regarding whether it can safely/for sure reset the machine. By accident it just hold the resources to do on some, not on all, machines.

The only component that knows for sure how to reset the machine is the acpica component because it parses the ACPI tables and obtaines there the way how to do it for the given machine.

I think it is technical of no good to let the platform driver poke on I/O ports without knowledge, but we may of course keep it.

> IMO requesting the I/O ports in the reset module of platform_drv only does no harm.

The only downside, from my point of view is, that the I/O port is then potentially not available to components actual really having a need for it, like acpica, e.g to re-configure devices according to findings in the acpi tables - but I have to admit that I have not seen this so far.

reet-:
@alex-ab I tested the current state on Muen and both `pci` and `vbox4_win7_32` scenarios work. Great!

I had to revert the changes you did in `platform_drv.inc` though, otherwise I get:

```
[init -> test-pci] --- Platform test started ---
[init -> acpi_drv] Error: ROM-session creation failed (ram_quota=6144, cap_quota=3, label="platform_info")
[init -> acpi_drv] Error: Could not open ROM session for "platform_info"
```

Should we enable `acpi_drv` for `x86 && !muen` only?

alex-ab:
@reet-: yes, I forgot of course to adjust for muen. afd36b4fbe46c960d856db7773e7724a5f4cf0f0 should help, does it ?

reet-:
@alex-ab Thanks for preparing the fixup. One part is missing though, please consider https://github.com/codelabs-ch/genode/commit/b0667df29fe9fbc76316d3ce61c77576206dcc80.

alex-ab:
@reet-: thanks, added.

reet-:
As requested by @chelmuth I tested @alex-ab's  [issue_2547_staging](https://github.com/alex-ab/genode/tree/issue_2547_staging) branch on qemu using nova and base-hw kernels with (rev 4c75c8c) and without (rev a988427) the mmconf patches. The results are equal.

alex-ab:
Thanks, for your efforts. However, I fear the issue_2547_staging (12/2017) branch is not merge-able because it does not apply to current staging. Some effort for re-merge and test on qemu and native hardware for _all x86 kernels_ is required.

reet-:
Yes sure. But before investing more time, the remaining open issues should be addressed (i.e. what to do with the reset functionality and also some problems with pistachio, as one commit is marked as HACK). Don't you agree?

alex-ab:
The commit message for pistachio must just be changed, it is actually no hack (see above):

> The remaining issue is pistachio, where @chelmuth has some ideas?
>
> @chelmuth: We discussed offline that @alex-ab's patch is correct as the KIP and UTCB regions are virtual and the > backing RAM is allocated by the kernel.

alex-ab:
Regarding the reset feature I explained my stand. Beside that there has been no (nor sound) technical discussion nor answers to keep it.

reet-:
Thanks @alex-ab for the information. So let's wait if someone speaks up regarding the reset issue. Otherwise I would rebase, re-test and re-submit this patchset.

nfeske:
I agree with @alex-ab that the reset handling should not be part of the platform driver. On the other hand, I share @chelmuth's sentiment that this feature should be available without running acpica. For example, in the sculpt scenario, I use it quite frequently and find it very convenient. I would definitely miss it if it vanished.

To reconcile both positions, I suggest to provide the reset functionality in another way. The lowest-hanging fruit would be to extend the PS/2 driver with the option to listen to the "system" ROM module and trigger a reset at the ps2 controller once the state changes to "reset". I am aware that this might not work for all platforms. But maybe we are lucky that it works at least for the ones we use? For a more robust way, we could - down the road - add a "reset" component to the drivers subsystems that issues a reset via PCI (I guess - not 100% sure about the other mechanisms) when the system ROM enters the reset state. This component would be a client of the platform driver. With this in place, both mechanisms would concurrently attempt to reset the machine. One would most certainly work. ;-)

To bring this issue forward, I suggest to try the PS/2 method first as this sounds like the easiest and least disruptive way.

chelmuth:
I agree with @nfeske and, thus, reset may vanish from x86/platform_drv.

reet-:
I rebased the [platform_drv-mmconf](https://github.com/codelabs-ch/genode/tree/platform_drv-mmconf) branch onto the current staging branch (e79ce5a03).

chelmuth:
I merged the commit series. Has anyone hints why the following is logged on Qemu (`make -C build/x86_64 run/ahci_bench KERNEL=nova`)?
```
[ 0] sys_assign_pci: Invalid Hint (0x18)
[init -> platform_drv] Error: ahci_drv -> : assignment of PCI device 0:3.0 failed phys=0xb0018000 virt=0x1000
```

alex-ab:
In case of Qemu this is expected. sys_assign_pci doesn't find a DMAR unit for this device.

alex-ab:
Unfortunately, the hack for the too short MMCONF/ECAM memory information broke the good machines. Please re-check and consider for staging. @chelmuth: https://github.com/alex-ab/genode/commit/b0d39b10ce6c7a4536547fde3f4f912d9e16fd09

https://github.com/alex-ab/genode/commit/7f9d154fe27fac4805387b35d602e7ab9a1f6b8c  fixes boot errors for ARM boards.

alex-ab:
@chelmuth: please add https://github.com/alex-ab/genode/commit/8fd7448474ec139156f470d51478cec7ee3d13c3 to staging

For Fiasco.OC and Fiasco another quirk would be required. Core in both cases gets stuck if the kernel can't map the I/O memory locally in core. Either we would have to change the generic Genode core code to cope with such a situation or we would have to make sure that the I/O memory regions for Fiasco.OC and Fiasco are actually accessible by Genode core. https://github.com/alex-ab/genode/commit/efbdab26628ac689f2f7f13932eb1842f9db15ea showcases it for Fiasco.OC (solely tested on x86 ...)

chelmuth:
I merged the fixup but left the Fiasco issue for tomorrow... maybe we could fix it in sigma0.

chelmuth:
I instrumented Fiasco with
```diff
diff --git a/repos/base-fiasco/src/core/io_mem_session_support.cc b/repos/base-fiasco/src/core/io_mem_session_support.cc
index 123b1e8..4b84a64 100644
--- a/repos/base-fiasco/src/core/io_mem_session_support.cc
+++ b/repos/base-fiasco/src/core/io_mem_session_support.cc
@@ -82,10 +82,18 @@ addr_t Io_mem_session_component::_map_local(addr_t base, size_t size)
 		                        &dw0, &dw1,
 		                      L4_IPC_NEVER, &result, &tag);
 
+		if (request == SIGMA0_REQ_FPAGE_IOMEM)
+			log("map I/O request"
+			    , " base=",       Hex(base)
+			    , " offset=",     Hex(offset)
+			    , " local_base=", local_base
+			    , " size=",       Hex(size)
+			    , " err=", err, "/", !l4_ipc_fpage_received(result)
+			    );
 		if (err || !l4_ipc_fpage_received(result)) {
 			error("map_local failed err=", err, " "
 			      "(", l4_ipc_fpage_received(result), ")");
-			return 0;
+//			return 0;
 		}
 
 		offset += 1 << page_size_log2;
diff --git a/repos/base-fiasco/src/core/platform_pd.cc b/repos/base-fiasco/src/core/platform_pd.cc
index a972c68..313e387 100644
--- a/repos/base-fiasco/src/core/platform_pd.cc
+++ b/repos/base-fiasco/src/core/platform_pd.cc
@@ -241,9 +241,14 @@ void Platform_pd::flush(addr_t, size_t size, Core_local_addr core_local_base)
 	using namespace Fiasco;
 
 	addr_t addr = core_local_base.value;
-	for (; addr < core_local_base.value + size; addr += L4_PAGESIZE)
+	for (; addr < core_local_base.value + size; addr += L4_PAGESIZE) {
+		log("unmap "
+		    , " addr=", Hex(addr)
+		    , " size=", Hex(L4_PAGESIZE)
+		    );
 		l4_fpage_unmap(l4_fpage(addr, L4_LOG2_PAGESIZE, 0, 0),
 		               L4_FP_FLUSH_PAGE);
+	}
 }
 
 Platform_pd::Platform_pd(Allocator *, char const *, signed pd_id, bool create)
```
and got a strange pattern where the region is I/O mapped twice. While the first mapping works as expected, the second fails, but only for the physical page 0xe1000000. @alex-ab maybe you can spot the error in the attached log?
[log.txt](https://github.com/genodelabs/genode/files/1841589/log.txt)

chelmuth:
The fiasco bug can be preproduced on Qemu by
```diff
diff --git a/repos/os/src/drivers/platform/spec/x86/pci_session_component.h b/repos/os/src/drivers/platform/spec/x86/pci_session_component.h
index e858c55..39bb292 100644
--- a/repos/os/src/drivers/platform/spec/x86/pci_session_component.h
+++ b/repos/os/src/drivers/platform/spec/x86/pci_session_component.h
@@ -156,6 +156,7 @@ class Platform::Pci_buses
                {
                        Config_access c(pciconf);
                        scan_bus(c, heap);
+                       scan_bus(c, heap, 0xff);
                }
 
                /**
@@ -883,7 +884,7 @@ class Platform::Root : public Genode::Root_component<Session_component>
 
                                Device_config const bdf_first(bdf_start);
                                Device_config const bdf_last(bdf_start + func_count - 1);
-                               addr_t const memory_size = 0x1000UL * func_count;
+                               addr_t const memory_size = 0x1100000UL;//0x1000UL * func_count;
 
                                /* Simplification: Only consider first config space and
                                 * check if it is for domain 0 */
```

chelmuth:
I feel we're on the finishing line. The last remaining imperfection is that platform_drv.inc now whines on base-linux with
```
init -> acpi_drv] Error: ROM-session creation failed (ram_quota=6144, cap_quota=3, label="platform_info")
[init -> acpi_drv] Error: Could not open ROM session for "platform_info"
Warning: no io_mem support on Linux (args="base=0xe0000, size=0x20000, wc=no, diag=0, label="init -> acpi_drv -> ", ram_quota=466, cap_quota=2")
[init -> acpi_drv] Segmentation fault (signum=11), see Linux kernel log for details
```
because acpi_drv and platform_drv are not required/usable on this platform.

alex-ab:
@chelmuth: 1401f8a81d5dadb7e798cc5d7bfc7fdc19b5d00d helps for me

PR #2550, opened 2017-10-26
RNDIS support

senier:
This branch adds support for Microsofts proprietary RNDIS protocol. With these changes it is possible to use the network sharing ("tethering") features provided by many Android devices. The driver was tested using the following devices:

* Samsung Galaxy S5
* HTC One A9
* Motorola Moto E

Many other devices should be supported out of the box. Using an ASIX AX88179 Gigabit Ethernet adapter, I confirmed that the existing USB Ethernet device driver still works.

As a side effect, I had to port the cdc_ether driver (the open protocol alternative phone vendors should be using) which the RNDIS driver is based on. Unfortunately I have no devices supporting CDC. Hence, while enabled, cdc_ether must be considered untested for now.


chelmuth:
I merged your commit series. Thanks for the neat feature! Could you please check as I had to fiddle with the port hashes.

senier:
Just tested staging (dd0271a7092aba167b6ad4a5af32367d2147c24e). Still works as expected. Thank you!

chelmuth:
Thanks.

PR #2572, opened 2017-11-10
dde_linux wifi: added firmware for Intel Wireless 3160

MartijnVerschoor:
Successfully tested on HP840 G2 with Intel® Dual Band Wireless-AC 3160 802.11 ac (1x1).


MartijnVerschoor:
Thanks!

PR #2597, opened 2017-12-01
acpica: introduced work around to fix shutdown for HP840 machines

MartijnVerschoor:
HP840 machines (g2/g3) call the unimplemented AcpiOsStall callback
during software shutdown. This drives acpica into an infinite loop which
prevents the shutdown. Instead, it is better to print an error message and 
continue.

### DIFF
--- a/repos/libports/src/lib/acpica/osl.cc
+++ b/repos/libports/src/lib/acpica/osl.cc
@@ -266,7 +266,10 @@ UINT64 AcpiOsGetTimer (void)
 	FAIL(0)
 
 void AcpiOsStall (UINT32)
-	FAIL()
+{
+//	FAIL()
+	Genode::error(__func__, " ");
+}
 
 ACPI_STATUS AcpiOsReadMemory (ACPI_PHYSICAL_ADDRESS, UINT64 *, UINT32)
 	FAIL(AE_BAD_PARAMETER)


alex-ab:
@MartijnVerschoor: may you please try commit 313c067264dc84b149c60497d4376e9ac4b74ff2 instead of yours ?

MartijnVerschoor:
@alex-ab, perfect! Thanks for providing the correct implementation of the unimplemented function so quickly.

I successfully tested this on HP840 G2/G3, Lenovo x250. Note that I have not seen the log message roll by. Probably because of the shut-down.

alex-ab:
Nice, that it works. Yes, it seems that the serial data gets not out of the machine quickly enough before the shutdown (due to buffering somewhere maybe ?). At the end, it is just a diagnostic message and you wanted to switch off/reset the machine anyway ;-) .

MartijnVerschoor:
Right, I just wanted to inform you that I have not been able to test the correct implementation of the AcpiOsStall callback function. I just checked that acpi shutdown works.

Shall I close this issue (PR) now or should I wait until the commit is pushed to staging?

alex-ab:
Just leave open, it will be closed automatically if the commit gets mastered (due to the ```Fix #number``` in the commit)

MartijnVerschoor:
Ok, thanks for the instruction.

PR #2607, opened 2017-12-11
base: separate toolchain and architecture

jklmnn:
This commit separates the `CROSS_DEV_PREFIX` into a toolchain and an architecture specific part. This makes it possible to use a custom toolchain in etc/tools.conf by overwriting `TOOLCHAIN_PREFIX` without the need to check for all architectures.
This is backwards compatible to it should not break any custom values of `CROSS_DEV_PREFIX`.

### DIFF
--- a/repos/base/etc/tools.conf
+++ b/repos/base/etc/tools.conf
@@ -18,18 +18,23 @@
 # We handle all architectures that may be specified as SPEC argument to the
 # package build tool.
 #
+
+TOOLCHAIN_PREFIX ?= /usr/local/genode-gcc/bin/genode-
+
 ifeq ($(filter-out $(SPECS),x86_32),)
-CROSS_DEV_PREFIX ?= /usr/local/genode-gcc/bin/genode-x86-
+TOOLCHAIN_ARCH ?= x86-
 endif
 ifeq ($(filter-out $(SPECS),x86_64),)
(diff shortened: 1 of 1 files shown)


nfeske:
The recommended way to use a custom tool chain is to place a custom `tools.conf` file in your `<build-dir>/etc/` directory. Since the build directory is architecture-specific anyway, each `<build-dir>/etc/tools.conf` can define an architecture-specific `CROSS_DEV_PREFIX`, can't it?

Note that in `base/mk/global.mk`, both the `tools.conf` as found in the repositories and the `tools.conf` as found in the build directory are included. So the version in your build directory may be as simple as:

```
CROSS_DEV_PREFIX := /path/to/your/custom/toolchain/prefix-architecture-
```

The checks in `base/etc/tools.conf` are merely meant to cover the default case of using the official tool chain. I am unable to see any practical benefit of your suggested change. But maybe I'm missing something?

The problem I see with splitting `CROSS_DEV_PREFIX` into pieces is that all tools (e.g., the depot and autopilot tools) must then also support this flexibility to maintain the current consistency of the tools. I would prefer to avoid that.

jklmnn:
I forgot that the build directory is architecture dependent so it doesn't make sense to provide a generic tool chain in etc/tools.conf. Since there are further issues with consistency I would just close this PR.

nfeske:
Thanks for the reconsideration.

PR #2656, opened 2018-01-25
pthread: Implement essential pthread_rwlock_* operations

senier:
Readers/writer lock implementation.

### DIFF
--- a/repos/libports/src/lib/pthread/rwlock.cc
+++ b/repos/libports/src/lib/pthread/rwlock.cc
@@ -1,0 +1,155 @@
+/*
+ * \brief  POSIX readers/writer lock (rwlock) implementation
+ * \author Alexander Senier
+ * \date   2018-01-25
+ *
+ */
+
+/*
+ * Copyright (C) 2018 Componolit GmbH
+ *
+ * This file is part of the Genode OS framework, which is distributed
+ * under the terms of the GNU Affero General Public License version 3.
+ */
+
+/* Genode includes */
+#include <base/log.h>
+#include <base/lock.h>
+#include <base/lock_guard.h>
+#include <base/thread.h>
+
+/* Libc includes */
+#include <errno.h>
+#include <pthread.h>
+
+/*
+ * A reader-preferring implementation of a readers-writer lock as described
+ * in Michael Raynal, "Concurrent Programming: Algorithms, Principles, and
+ * Foundations", ISBN 978-3-642-32026-2, page 75
+ */
+
+extern "C" {
+
+	/*
+	 * This class is named 'struct pthread_rwlock' because the 'pthread_rwlock_t'
+	 * type is defined as 'struct rwlock*' in 'sys/_pthreadtypes.h'
+	 */
+	struct pthread_rwlock
+	{
+		private:
+
+			Genode::Thread *_owner;
+			Genode::Lock _nbr_mutex {};
+			Genode::Lock _global_mutex {};
+			int _nbr = 0;
+
+		public:
+
+			void rdlock()
+			{
+				Genode::Lock_guard<Genode::Lock> guard(_nbr_mutex);
+				++_nbr;
+				if (_nbr == 1) {
+					_global_mutex.lock();
+				}
+			}
+
+			void wrlock()
+			{
+				_global_mutex.lock();
+				_owner = Genode::Thread::myself();
+			}
+
+			int unlock()
+			{
+				Genode::Lock_guard<Genode::Lock> guard(_nbr_mutex);
+
+				/* Read lock */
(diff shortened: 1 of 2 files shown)


repos/libports/src/lib/pthread/rwlock.cc
```diff
@@ -0,0 +1,156 @@
...
+			Genode::Lock_guard<Genode::Lock> guard(_nbr_mutex);
+			_nbr--;
+			if (_nbr == 0) {
+				_global_mutex.unlock();
+			}
+		}
+
+		void _begin_write() { _global_mutex.lock(); }
+
+		void _end_write() { _global_mutex.unlock(); }
```
chelmuth: From beginning of class until here the indentation misses one level (of tabs).

repos/libports/src/lib/pthread/rwlock.cc
```diff
@@ -0,0 +1,156 @@
...
+
+		public:
+			int rdlock()
+			{
+				if (_state != RWLOCK_UNLOCKED) {
+					Genode::warning("Read lock on locked rwlock (state=", int(_state), ")");
```
chelmuth: Is this warning really needed? Wouldn't produce too much noise if there's actually contention? (consider also other warnings below)

repos/libports/src/lib/pthread/rwlock.cc
```diff
@@ -0,0 +1,156 @@
...
+		if (attr != nullptr) {
+			Genode::error(__func__, " attr was not NULL");
+			return -1;
+		}
+
+		*rwlock = new struct pthread_rwlock();
+		          return 0;
+	          }
+
+	          int pthread_rwlock_destroy(pthread_rwlock_t *rwlock)
```
chelmuth: Another indenation error.

repos/libports/src/lib/pthread/rwlock.cc
```diff
@@ -0,0 +1,156 @@
...
+			}
+
+			int wrlock()
+			{
+				if (_state != RWLOCK_UNLOCKED) {
+					Genode::warning("Write lock on locked rwlock (state=", int(_state), ")");
+					return -1;
+				}
+				_state = RWLOCK_WRITE;
+				_begin_write();
```
chelmuth: I don't know the article you refer to but couldn't there be preemption before each statement in `rdlock()` and  `wrlock()`? So, how is ensured that `_state` matches the actual lock owner (`READ` or `WRITE`)? Is this property important because it is used on `unlock()`?

repos/libports/src/lib/pthread/rwlock.cc
```diff
@@ -0,0 +1,156 @@
...
+				_state = RWLOCK_READ;
+				_begin_read();
+				return 0;
+			}
+
+			int wrlock()
+			{
+				if (_state != RWLOCK_UNLOCKED) {
+					Genode::warning("Write lock on locked rwlock (state=", int(_state), ")");
+					return -1;
```
chelmuth: Acoording to http://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_rwlock_wrlock.html `errno` should be set to `EBUSY`.

senier:
The `_state` variable was meant to correlated `wrlock()`/`rwunlock()` and `rdlock()`/`rdunlock()` pairs. As you spotted, this was bogus. (And, the name `_state` is misleading).

Here is the problem: The pthreads interface defines only `rwlock()`, `rdlock()` and `unlock()`, i.e. `unlock()` is used for unlocking both, read and writer locks. Unfortunately, all the (textbook) algorithms I found assume the presence of dedicated unlock operations.

The best I can think of is to make `_state` thread-local as done with pthread_self in thread.cc. But I'm open for other ideas.

chelmuth:
No idea, sorry. Please don't miss my other comments.

senier:
Reworked the pthread_rwlock implementation. The owner of the writer lock is now tracked so we can select the right unlock operation using the thread ID (which is thread-local).
I also ported parts of the respective [NPTL test cases](https://github.com/senier/componolit/tree/pthread_rwlock_tests/src/test/concurrency/pthread). May I add those?

chelmuth:
I like the revised implementation. @senier would you mind to add a copyright header and add it to the library mk file?

senier:
Thanks for reviewing!

I added a copyright header. What else do you want me to put into the library mk file? (the new `rwlock.cc` was added to `SRC_CC` in `repos/libports/lib/mk/pthread.mk` - anything else I'm missing?)

chelmuth:
Sorry, seems I was confused yesterday. Merged your commit to staging. Thanks!

PR #2675, opened 2018-02-10
Noux packages updates

N8Fear:
Hi,
here is the first bunch of updates I asked about on the mailing list.
In most cases the version bump is relatively straigt forward. I tested all changes on qemu/nova witwith a modified version of 'run/noux_bash', which for that reason I included as well.
In case of bash my version but is a little bit hacky as the documentation (especially man2html) caused linking errors due to mixing system gcc and the genode toolchain. Therefore I opted to just disable it, which works fine. Alternatively I could try to pour in some more time and try to find a better solution.
I'll try to get other stuff updated in the future.
If you have comments feel free to contact me.


nfeske:
Cool!

While reviewing the patches, the following questions popped into my mind:

* You changed the download location of less and bumped the version to 529. According to the official GNU site http://ftp.gnu.org/gnu/less/, the latest version is 487, which is also stated at http://www.greenwoodsoftware.com/less/index.html. Shouldn't we better stick with the official release version instead of a beta version?

* Your update of bash changes the check_dev_tty.patch in a non-intuitive way. The original comment by Christian Prochaska is gone (but is still applies, doesn't it?). You added the line `malloc_set_register (1);    /* XXX - change to 1 for malloc debugging */` which I cannot relate to the purpose of the patch. Wouldn't it be better to preserve the comment and add a note behind the intention of the `malloc_set_register(1)` call? Regarding your question above, I think it's perfectly fine to disable the docs for bash.

I'd like to leave out the noux_bash.run patch to keep the run script as small as possible. The changes can easily be applied manually whenever needed. For the background, I regularly use this run script as a quick way to test noux with an interactive shell when working on basic noux mechanisms like pipes and fork. Hence, building all these packages would slow down this work flow.

N8Fear:
Hello Norman,
I've changed less to version 487 and back to the GNU page. This was due to a misunderstanding on my part. You're right that the stable release is more preferable.
It's similar regarding the patch to bash (check_dev_tty): I tested some stuff unrelated to that "abusing" the existing patch and forgot to revert the change before committing. I've done so now and tested it again.
In case of noux_bash.run: I never intended to get this merged, but thought it may be helpful for testing/reviewing and so I added it here. I've changed the wording of the commit message to hopefully make that clear... ;)

nfeske:
Wow that was quick! Thanks a lot for addressing my remarks. I merged your new commit series to our staging branch. Let's see what our automated tests have to say. ;-)

alex-ab:
Closing because merged to master.

PR #2688, opened 2018-02-20
Muen port update to v0.9

Kensan:
These changes update the Muen port to the latest release version (0.9) allowing to drop the patch which was integrated upstream. The documentation has been updated to reflect the necessary changes.

Note: the GPL 2017 toolchain is now required and as the debug output format has changed the _mulog-subject.py_ script must be updated on autopilot instances.

We ran the changes on our autopilot instance and the results look good. I am aware that you are busy testing Sculpt (which is very exciting btw!) but it I hope you find time to include the commits so the port update could make it into the 18.02 release. :)


nfeske:
Thank you for submitting the update!

As far as I can see, this change won't affect sculpt at all. @chelmuth I have no reservations against merging it to staging. However, @Kensan please note that we cannot dedicate much time on troubleshooting (should there by any problem with the new Muen version) prior the release.

Kensan:
Thanks!

> However, @Kensan please note that we cannot dedicate much time on troubleshooting (should there by any problem with the new Muen version) prior the release.

Yes, we are aware and will do additional autopilot runs & testing on our end.

chelmuth:
I merged the commits as they do not depend on #2547, which is still pending.

chelmuth:
Merged into master by 9e4ff8c196c1c588bee4eb00cd959820e3d923aa

PR #2721, opened 2018-03-17
Add missing libc ABI symbols

senier:
Those were missing to build FreeBSD's `rm`.

### DIFF
--- a/repos/libports/lib/symbols/libc
+++ b/repos/libports/lib/symbols/libc
@@ -195,6 +195,7 @@ fseeko T
 fsetpos T
 fstat T
 fstatat T
+fstatfs T
 fstatvfs T
 fsync T
 ftell T


chelmuth:
I agree that it makes sense to add `fstatfs` to the API, but I'd like to question we should or could support `undelete` and `lchflags`. These functions seem to be FreeBSD-specific and are not enabled by the current libc backend. What happens if they are called?

senier:
You are right, I should have analyzed the source more carefully. `lchflags` is used to remove append-only and immutable flags from files before removal. The `undelete` call is for even more esoteric reasons - it is used in cases where a file has been removed by a "whiteout" in an upper layer of a union file system. This also appears to be the only uses case for `undelete` on FreeBSD.

I can mock those functions locally instead of using the libc dummy implementations. See my update.

PR #2748, opened 2018-04-09
ada: runtime and library support

jklmnn:
I added global Ada runtime and library support.

The runtime is not anymore inside the program dir but can be set with the `ADA_RTS` variable. The `ada` library builds a (currently) minimal runtime from the gcc sources that come with Genode (to stay consistent with the used compiler) and sets the runtime path accordingly. It is build as a shared library `ada.lib.so` which needs to be added to the build files.

Library support worked out of the box since `INC_DIR` also works for gnatmake and ads files. I only splitted the existing Ada test into a program and a library.


repos/base/lib/import/import-ada-lib.mk
```diff
@@ -0,0 +1,2 @@
+
+INC_DIR += $(REP_DIR)/src/test/ada-lib
```
chelmuth: I'd prefer this called *test-ada* as it is used in the test case only.

repos/base/lib/mk/ada-lib.mk
```diff
@@ -0,0 +1,6 @@
+LIBS += base ada
+SRC_ADA += add_package.adb
+
+include $(REP_DIR)/lib/import/import-ada-lib.mk
+
+vpath %.adb $(REP_DIR)/src/test/ada-lib
```
chelmuth: Same here.

repos/base/mk/generic.mk
```diff
@@ -7,7 +7,7 @@
 # Collect object files and avoid duplicates (by using 'sort')
 #
 SRC_O  += $(addprefix binary_,$(addsuffix .o,$(notdir $(SRC_BIN))))
-SRC     = $(sort $(SRC_C) $(SRC_CC) $(SRC_ADA) $(SRC_RS) $(SRC_S) $(SRC_O))
+SRC     = $(sort $(SRC_C) $(SRC_CC) $(SRC_ADA) $(SRC_RS) $(SRC_S) $(SRC_O) $(SRC_ADS))
```
chelmuth: Just nitpicking: It may make sense to add `SRC_ADS` right behind `SRC_ADA`.

repos/libports/lib/mk/ada.mk
```diff
@@ -0,0 +1,24 @@
+
+RTS_DIR = $(BUILD_BASE_DIR)/var/libcache/ada
```
chelmuth: If this always equal to `ADA_RTS` above, why not use one variable?

repos/base/lib/import/import-ada-lib.mk
jklmnn: I would call this `test-ada-lib` then since `test-ada` is already the name of the test component.

repos/base/mk/generic.mk
jklmnn: Maybe one could replace `SRC_ADA` with `SRC_ADB` also since this is distinguished anyway now?

repos/base/lib/import/import-ada-lib.mk
chelmuth: Sure.

repos/base/mk/generic.mk
chelmuth: Sounds good to me, do ahead.

senier: `SRC_ADS` and `SRC_ADB` definitely make sense. We should change it and adapt the few Ada components now.

chelmuth:
I reintegrated the commits today and added a fixup with 015d60d3bd54faae4edde7395fee06c55a1b20f4, which adapts naming and moves the test case to libports as it depends on the ada runtime. Please review the changes and also the commit message, which is a slightly changed adaption of the PR description and will be squashed to the original commit.

Please also review *repos/libports/src/test/ada/README* as it seems somewhat outdated to me.

jklmnn:
I looked into the fixup and moving it to libports seems a good idea to me. I'm also fine with the commit message.

The current README is not necessarily outdated but a bit incomplete. I would add
"To use Ada the `ada` library is required as it provides the needed runtime." Aside from that I don't see any missing things.

chelmuth:
Okay, I reviewed the README and did some other cleanups in a40dae74ae64e0a28576c83959c9fb4bf329828f. From my perspective this wraps up the PR.

chelmuth:
Sorry, I was to hasty. We still have a regression when building for x86_32.
```
  Library ada
    COMPILE  system.o
    MERGE    ada.lib.so
/usr/local/genode-gcc/bin/genode-x86-ld: i386:x86-64 architecture of input file `system.o' is incompatible with i386 output
.../genode/genode_staging.git/repos/base/mk/lib.mk:246: recipe for target 'ada.lib.so' failed
make[3]: *** [ada.lib.so] Error 1
```

jklmnn:
https://github.com/genodelabs/genode/pull/2748/commits/0b473a1595b5a7ad95bc04abd93d87e8ff5593d1 should fix this.

chelmuth:
Sorry, I already added a201fe916d416dbabb81c1fd9147a9339e4ad8f5 to staging.

jklmnn:
No problem, thank you for fixing this.

jklmnn:
@chelmuth could you please add https://github.com/jklmnn/genode/commit/045e1ece47b1d9dae1a0734dc3ec75649de9f200 to this PR? This is required to build parts of the runtime that consist of both an ads and an adb file.

nfeske:
The last commit conflicts with current staging. @jklmnn Would you mind rebasing it to staging and fixing the alignment of the `=` characters while being at it?

jklmnn:
https://github.com/jklmnn/genode/commit/cfd5d87438eec4e1fb10de8f0f46e033088c3481 should apply on staging.

nfeske:
Cool! I merged it as 45adb4ab1a496362344bb3fc7d69e330860407ca to staging. Note that I had to change the commit message and keep it separate from the original commit because 1261c18ce97b46258c2e6fe0a46b72e356081884 went already to master.

PR #2764, opened 2018-04-13
sel4: update to version 9.0.0

N8Fear:
Hi,

I updated the seL4 port to use seL4 version 9.0.0. The required changes are relatively small and it seems to work fine.
I tested it on x86_64 and x86_32 with a bunch of scenarios (log, thread, demo, noux_terminal_fs and genode_org) and it seems to work fine (or at least as well as on current master: on x86_64 the heavier scenarios like demo seem to need adjustments concerning RAM and CAPS, but that's the same as on current master).
I also tested the build on wand_quad but due to lack of of hardware I couldn't test the function. Due to the kind of code changes I expect it to work anyways.


repos/base-sel4/src/core/platform.cc
```diff
@@ -402,6 +393,28 @@ void Platform::_init_rom_modules()
...
+
+				tsc_freq const * boot_freq = reinterpret_cast<tsc_freq const *>(reinterpret_cast<addr_t>(element) + sizeof(* element));
+
+				xml.node("hardware", [&] () {
+					xml.node("features", [&] () {
+						xml.attribute("svm", false);
+						xml.attribute("vmx", false);
+					});
+					xml.node("tsc", [&] () {
+						xml.attribute("freq_khz" , boot_freq->freq_mhz / 1000UL);
```
chelmuth: IMO the calculation should be `kHz=1000*MHz`, right?

N8Fear: You are correct. Must have been a brain lag on my part...

alex-ab:
Nice !

N8Fear:
Should I squash the fixup commit?

alex-ab:
You may do so. Your commits will be considered as soon as current staging branch settles down.

N8Fear:
Done.
No rush. I guess I'll stay for a while... ;)

N8Fear:
I've added the update to 9.0.1. There is a potentially breaking change for 64 bit arches:

>  * On 64-bit architectures, the `label` field of `seL4_MessageInfo` is now 52 bits wide. User-level programs
   which use any of the following functions may break, if the program relies on these functions to mask the
   `label` field to the previous width of 20 bits.
     - `seL4_MessageInfo_new`
     - `seL4_MessageInfo_get_label`
     - `seL4_MessageInfo_set_label`

As far as I can tell this doesn't matter to Genode, because for `seL4_MessageInfo_new` we use `seL4_MessageInfo_t` from the seL4 sources as type and for `seL4_MessageInfo_get_label` we use `_fault_type` which is of type `addr_t` and should be 64 bit wide on 64bit arches.
Because of that I think the version bump should not require any additional changes.

N8Fear:
Shall I close this, now that the commits are pulled?

N8Fear:
Thank you for merging it!

PR #2829, opened 2018-05-15
WIP: base-linux-hw

jklmnn:
This work in progress PR targets base-linux to be able to use native Genode drivers. It also enables Genode to boot on a bare Linux kernel and creating images for image/iso and image/uefi. To do this
```
KERNEL_RUN_OPT(linux) := --include power_on/linux --include log/linux
```
needs to be replaced with
```
KERNEL_RUN_OPT(linux)   := $(QEMU_RUN_OPT)
```
in etc/build.conf.
Also for some components to build `pci` needs to appended to repos/base-linux/etc/specs.conf

# Current state

## Linux kernel

The Linux kernel used is available at https://github.com/jklmnn/linux/tree/genode. Interrupts and MMIO get provided by the [hwio](https://github.com/jklmnn/linux/blob/genode/drivers/char/genode/hwio.c) module and the `platform_info` ROM gets provided by the [platform_info](https://github.com/jklmnn/linux/blob/genode/drivers/char/genode/platform_info.c).

## IO ports

The `IO_PORT` session works without any problems.

## IO memory

The `IO_MEM` session work for mapping and using memory but freeing it results in a segmentation fault. This prevents the acpi_drv from working. If `free_io_mem` is implemented with a dummy the driver works but obviously creates a memory leak.

## Interrupts

Interrupts are provided by calling a blocking read on the hwio device that returns on a triggered interrupt (similar to okl4).

## DMA

To provide DMA on base-linux the RAM session using mmaped files needs to replaced with an implementation that provides contiguous physical memory. The current implementation of the hwio kernel module already provides a configuration for this case but replacing the file used in Genodes RAM session with this file descriptor causes a segmentation fault in the child component.

## Architectures

Currently this setup only works for x86_64 because of the provided kernel image. A better solution would be to build the bzimage from a port of the Linux kernel with the correct compiler and architecture.

## Notes

The fb_drv fails to load due to a race condition. On qemu this can be mitigated by inserting  log messages before the call of `X86EMU_exec()` (and therefor slowing down the driver).

I hope I didn't forget anything. If something is unclear, don't hesitate to ask me.

TODO:
 - [ ] separate Linux kernel port
 - [x] distinction between base-linux and base-linux-hw
 - [ ] DMA support
 - [ ] Working deallocation of IO memory


tool/run/image/uefi
```diff
@@ -32,7 +33,7 @@ proc run_image { {unused ""} } {
 	exec dd if=/dev/zero of=[run_dir].partition bs=1M count=$disk_size 2>/dev/null
 	exec mkfs.vfat -n GENODE [run_dir].partition
 
-	# copy content to disk image
+        # copy content to disk image
```
nfeske: This might be an accident.

tool/run/image/uefi
```diff
@@ -20,10 +20,11 @@ proc run_image { {unused ""} } {
 
 	set run_size [expr [regsub {\s.*} [exec du -sm [run_dir]] {}]]
 
+
 	if {[image_uefi_size] > 0} {
 		set disk_size  [image_uefi_size]
 	} else {
-		set disk_size [expr $run_size + 1]
+		set disk_size [expr $run_size + 5]
```
nfeske: It would be good to know the rationale behind the number. (maybe as a comment?)

tool/run/image/uefi
```diff
@@ -20,10 +20,11 @@ proc run_image { {unused ""} } {
 
 	set run_size [expr [regsub {\s.*} [exec du -sm [run_dir]] {}]]
 
+
```
nfeske: Accident?

tool/run/run
```diff
@@ -940,6 +940,29 @@ proc build_core_image { modules } {
...
+    exec cp -f [run_dir]/genode/config [run_dir].config
+
+    set here [pwd]
+    cd [run_dir]
+    puts "generating initrd"
+    exec cp genode/initramfs init
+    exec mkdir tmp
+    set files "init\ntmp\ngenode\n"
+    append files [exec find genode -type f,l -printf "genode/%f\n"]
+    exec -ignorestderr echo $files | cpio -o -L -H newc > initrd
```
nfeske: Here, it would be good to wrap `cpio` into the `[installed_command]` function. See https://github.com/genodelabs/genode/commit/10485b0d39f2dc5d2a8a555449007055e7029324

repos/base-linux/src/core/linux/target.mk
```diff
@@ -23,6 +23,8 @@ SRC_CC        = main.cc \
                 ram_dataspace_factory.cc \
                 core_rpc_cap_alloc.cc \
                 io_mem_session_component.cc \
+		io_port_session_component.cc \
```
nfeske: whitespace: use spaces for alignment.

repos/base-linux/src/core/platform_services.cc
```diff
@@ -0,0 +1,39 @@
...
+
+/*
+ * Add x86 specific ioport service
+ */
+void Genode::platform_add_local_services(Rpc_entrypoint*,
+                                         Sliced_heap *sliced_heap,
+                                         Registry<Service> *local_services)
+{
+        if(!lx_iopl(3)){
+        	static Io_port_root io_port_root(core_env()->pd_session(),
```
nfeske: whitespace: use tabs for indentation, spaces for alignment. See https://genode.org/documentation/developer-resources/coding_style#Indentation

There are more occurrences of such whitespace issues. It would be nice of you to recheck.

tool/run/run
```diff
@@ -956,10 +956,20 @@ proc build_initrd { modules } {
...
+    set files "init\nhwio.ko\nplatform_info.ko\ntmp\ndev\ngenode\n"
     append files [exec find genode -type f,l -printf "genode/%f\n"]
     exec -ignorestderr echo $files | cpio -o -L -H newc > initrd
+    
+    #workaround because cpio fails to compress broken links sometimes
+    exec touch dev/platform_info
+    cd genode
+    exec ln -s ../dev/platform_info platform_info
+    cd ..
+    exec -ignorestderr echo "genode/platform_info" | cpio -o -A -H newc -O initrd
```
nfeske: Use `installed_command` for cpio

repos/base-linux/src/core/include/io_mem_session_component.h
```diff
@@ -26,6 +26,22 @@ namespace Genode {
 
 	class Io_mem_session_component : public Rpc_object<Io_mem_session>
 	{
+                private:
+
+                        enum {
+                            PAGE_SIZE = 4096,
+                            PAGE_SIZE_LN = 12
+                        };
+
```
nfeske: trailing whitespace

repos/base-linux/src/core/include/io_mem_session_component.h
```diff
@@ -26,6 +26,22 @@ namespace Genode {
 
 	class Io_mem_session_component : public Rpc_object<Io_mem_session>
 	{
+                private:
+
+                        enum {
+                            PAGE_SIZE = 4096,
+                            PAGE_SIZE_LN = 12
+                        };
```
nfeske: Better use existing definition at https://github.com/genodelabs/genode/blob/master/repos/base/src/include/base/internal/page_size.h

repos/base-linux/src/core/include/io_mem_session_component.h
```diff
@@ -26,6 +26,22 @@ namespace Genode {
...
+                        
+                        Range_allocator *_io_mem_alloc;
```
nfeske: Better use references instead of pointers

repos/base-linux/src/core/io_mem_session_component.cc
```diff
@@ -12,14 +12,60 @@
...
+
+#include <core_linux_syscalls.h>
 
 #include <io_mem_session_component.h>
 
 using namespace Genode;
 
+size_t Io_mem_session_component::get_arg_size(const char *args)
+{
+    size_t size = Arg_string::find_arg(args, "size").ulong_value(0);
```
nfeske: Indent with tabs

nfeske: variables should better be `const` values whenever possible, e.g., instead of `size_t size` use `size_t const size`.

repos/base-linux/src/core/io_mem_session_component.cc
```diff
@@ -12,14 +12,60 @@
...
+    addr_t base = Arg_string::find_arg(args, "base").ulong_value(0);
+
+    addr_t end = align_addr(base + size, PAGE_SIZE_LN);
+    base = base & ~(PAGE_SIZE - 1);
+    return (size_t)(end - base);
+}
+
```
nfeske: use two empty lines between method implementations, see https://genode.org/documentation/developer-resources/coding_style#Vertical_whitespaces

repos/base-linux/src/core/io_mem_session_component.cc
```diff
@@ -12,14 +12,60 @@
...
+addr_t Io_mem_session_component::get_arg_phys(const char *args)
+{
+    return Arg_string::find_arg(args, "base").ulong_value(0);
+}
+
+Cache_attribute Io_mem_session_component::get_arg_wc(const char *args)
+{
+    Arg a = Arg_string::find_arg("wc", args);
+    if(a.valid() && a.bool_value(0)){
```
nfeske: leave space around parenthesis

repos/base-linux/src/core/io_mem_session_component.cc
```diff
@@ -12,14 +12,60 @@
...
+        return WRITE_COMBINED;
+    }else{
```
nfeske: spaces around `else`

repos/base-linux/src/core/io_mem_session_component.cc
```diff
@@ -12,14 +12,60 @@
...
+                                                   Range_allocator *ram_alloc,
+                                                   Rpc_entrypoint  *ds_ep,
+                                                   const char      *args) :
+    _io_mem_alloc(io_mem_alloc),
+    _ds(get_arg_size(args), 0, get_arg_phys(args), get_arg_wc(args), true, 0),
+    _ds_ep(ds_ep)
+{
+    int _fd = -1;
+
+    _fd = lx_open("/dev/hwio", O_RDWR | O_SYNC);
```
nfeske: rename `_fd` to `fd` because it is not a private member but a local variable. Assign the return value of `lx_open` directly to `fd`. Turn `int fd` into `int const fd`.

repos/base-linux/src/core/io_mem_session_component.cc
```diff
@@ -12,14 +12,60 @@
...
+    lx_ioctl_iomem(_fd, (unsigned long)get_arg_phys(args), get_arg_size(args));
+    log("_fd ", _fd);
+    if(_fd > 0){
```
nfeske: spaces around parenthesis. Please check for further occurrences.

repos/base-linux/src/lib/base/region_map_mmap.cc
```diff
@@ -187,6 +187,9 @@ Region_map::Local_addr Region_map_mmap::attach(Dataspace_capability ds,
 		throw Region_conflict();
 	}
 
+        if(!ds.valid())
```
nfeske: space before opening parenthesis.

repos/base-linux/src/lib/base/region_map_mmap.cc
```diff
@@ -187,6 +187,9 @@ Region_map::Local_addr Region_map_mmap::attach(Dataspace_capability ds,
...
+            throw Invalid_dataspace();
```
nfeske: use tabs to indent

repos/base-linux/src/core/include/irq_object.h
```diff
@@ -0,0 +1,38 @@
...
+class Genode::Irq_object : public Thread_deprecated<4096>
+{
+
+    private:
+
+        Genode::Signal_context_capability _sig_cap;
+        Genode::Lock _sync_ack;
+        Genode::Lock _sync_bootup;
+        unsigned _irq;
+        int _fd;
```
nfeske: `_irq` and `_fd` should better be const

repos/base-linux/src/core/include/irq_session_component.h
```diff
@@ -25,32 +27,34 @@ namespace Genode {
 class Genode::Irq_session_component : public Rpc_object<Irq_session>,
                                       private List<Irq_session_component>::Element
 {
-	private:
+        private:
```
nfeske: use tabs to indent

repos/base-linux/src/core/include/irq_session_component.h
```diff
@@ -25,32 +27,34 @@ namespace Genode {
...
 
-		friend class List<Irq_session_component>;
+	    friend class List<Irq_session_component>;
```
nfeske: reindent

repos/base-linux/src/core/irq_session_component.cc
```diff
@@ -0,0 +1,113 @@
...
+
+#include <core_linux_syscalls.h>
+
+Genode::Irq_session_component::Irq_session_component (Genode::Range_allocator *irq_alloc, const char *args)
+:
+    _irq_number(Arg_string::find_arg(args, "irq_number").long_value(-1)),
+    _irq_alloc(irq_alloc),
+    _irq_object(_irq_number)
+{
+    /*
```
nfeske: remove commented-out code

nfeske:
Thanks for the commit series and the description!

I just reviewed the commits and have the following remarks. For minor things, I left comments at the individual commits.

*boot: boot Genode on Linux on UEFI* The commit contains the Linux kernel as a binary. Adding big binaries like that to the Genode repository is not a good idea. It inflates the git repository for everyone even though the binary is used by only a few. A better approach would be adding a port description that fetches the binary from a distinct place, like a separate repository.

*tool: add io mem support for linux* The binaries for the kernel modules should not be hosted in the Genode repository.

*boot: remove ACPI support from linux* should not be a distinct commit. In general, intermediate steps within the commit series should better be avoided by squashing them together. The same applies for *tool: linux: remove mouse and keyboard support*, and *tool: enable irq in hwio*.

To avoid me nagging you too much about coding-style issues, you may give the `tool/beautify` tool a try. Admittedly, I haven't used it for a long while now and I'm not sure how well that works, but was meant to check (and optionally fix) many style issues.

jklmnn:
Thanks for your review! While it is over a year ago that I have used the beautify tool the last time I think it'll help me ;)

About the Linux kernel binary: I'd like to remove it completely from the repository and build it from a port of https://github.com/jklmnn/linux/tree/genode. This solves not only the problem with binaries in git but also the architecture problem. Yet I'm not sure how to build a complex project like that inside Genodes build system. Only building the bzImage doesn't take that long (at least not longer than Qt) so I think thats feasible.

nfeske:
The most straight-forward way would be following the patterns of the other kernels. E.g., for sel4, the Genode build system just invokes sel4's regular bui/ld system. It is useful to do this wrapping from a library build target. So the output of the third-party build system stays nicely confined in a directory under var/libcache.

For reference, the target.mk file for the sel4 kernel looks like this: https://github.com/genodelabs/genode/blob/master/repos/base-sel4/src/kernel/sel4/target.mk

As you can see, it is not much more than a library dependency and a rule for creating a symlink in <build-dir>/bin/ that points to the kernel image. The interesting things happen in the library description file: https://github.com/genodelabs/genode/blob/master/repos/base-sel4/lib/mk/spec/x86_64/kernel-sel4.mk Here, the 3rd-party build system is invoked as a side effect of building the library. As the library is build from within var/libcache/kernel-sel4/, all files generated by the 3rd-party build system end up at this place.

Another (more complicated) example of the same pattern is building the Fiasco.OC kernel. The rules are located at https://github.com/genodelabs/genode/blob/master/repos/base-foc/lib/mk/kernel-foc.inc

The run tool then only picks up the kernel's binaries from the build directory.

tool/run/run
jklmnn: Was this already available in 17.08? Grepping through the repo doesn't find any occurence of this string.

repos/base-linux/src/core/include/irq_object.h
jklmnn: `_fd` cannot be const since it needs to be set in `Genode::Irq_object::_associate()`.

tool/run/run
nfeske: It was previously called `check_installed`. As this name is a bit misleading, we have recently renamed to `installed_command`. The corresponding issue is https://github.com/genodelabs/genode/issues/2853

repos/base-linux/src/core/include/irq_object.h
nfeske: alright. Thanks for clarifying.

tool/run/run
jklmnn: For now I would use `check_installed` and update it once this branch is rebased to the current version of Genode. I want to keep a working state in this branch.

jklmnn:
I added two fixup commits: I reviewed each file with `tool/beautify`, I hope I didn't forget something.

nfeske:
Cool. This looks very nice now. Regarding the copyright, the same remark applies as for the other pull request. Would you mind to add Genode Labs in addition to Componolit. E.g., like it's done here: https://github.com/genodelabs/genode/blob/master/repos/os/include/rtc_session/rtc_session.h

nfeske:
For further review, I'd very much appreciate a new branch rebased on the current genode/master, with intermediate commits and binaries removed. @jklmnn No need to hurry.

jklmnn:
I have rebased [base_linux_hw_r1](https://github.com/jklmnn/genode/tree/base_linux_hw_r1) to the current master and removed all binaries. Due to that you need to supply your own vmlinux to tool/boot (until I created a port for the Linux kernel) if you want to run the scenario.
Is there any way to let this PR track the new branch without overwriting the old one?

nfeske:
That was quick! Thanks a lot. The new branch is much better digestible for review.

Am I right that to try out the branch, I need the custom Linux kernel plus your kernel module? Are you planning to install both using the same port? That would be quite convenient.

Regarding the distinction of the hosted version versus the hw version of base-linux, we need to be careful to not break too many tests. I suspect that there are quite a few run scripts checking for the availability of qemu to distinguish test platforms. I'll consult with @chelmuth next week.

BTW, the story about the file headers applies to this branch as well. ;-)

> Is there any way to let this PR track the new branch without overwriting the old one?

I don't know if pull requests can be changed like that. As a workaround, maybe you could push your old branch under a different name (to preserve it) and force-push the new one under the original name?

jklmnn:
The kernel module is included in the kernel itself since this kernel version does not have loadable module support. You can either take it from https://github.com/jklmnn/linux/tree/genode and build it manually or just extract the latest binary of the old branch.
Due to the modules being built into the kernel there won't be a separate port.

I did not come to and final conclusion about the distinction of both base-linux version. The most important requirement is that I don't want to have any code copied or being redundant (I assume thats important for you too). Aside from that feel free to use any solution that fits best. I'm curious about your ideas ;)

Well the file headers that are already existing should have the correct format and information since I fixed them in the last rebase. Yet there are probably files without any headers at all, I'll take a look at that.

Okay, so I will create a backup branch called `base_linux_hw_r0_orig` and force push the current version on this one.

jklmnn:
I added a todo list to the first commit for issues that still need to be resolved. While you are free to merge earlier a state without these issues fixed is hardly useful.

nfeske:
I think I'll defer the merging until the feature can be used without jumping through hoops.

Regarding the distinction between both Linux variants, I am thinking in the direction as applied for executing NOVA in Qemu vs. real hardware. Here we use the RUN_OPT mechanism of the run tool, which can be tweaked in the build directory's `etc/build.conf`. However, I don't know how to name the run parameters in a good way, yet.

jklmnn:
Actually changing the `RUN_OPT` is what I currently do to distinguish between both version (see the initial comment). Only image building does not check this option so the images get build always.

jklmnn:
I updated the work on this PR to Genode 19.08. The linux variant can now be selected by setting the correct `BOARD` variable. `linux` will build the usual base linux and `pc` will build base linux on a bare linux kernel and run it inside Qemu.

jklmnn:
An alternative to the current custom kernel module could be [vfio](https://www.kernel.org/doc/html/latest/driver-api/vfio.html).

tool/run/image/uefi
```diff
@@ -19,7 +19,7 @@ proc run_image { {unused ""} } {
 	if {[image_uefi_size] > 0} {
 		set disk_size  [image_uefi_size]
 	} else {
-		set disk_size [expr $run_size + 1]
+		set disk_size [expr $run_size + 20]
```
chelmuth: What's your rationale to add 20M to `disk_size` here? IMO it's better to use the `--image--uefi--size` parameter for custom configurations.

tool/run/boot_dir/linux
```diff
@@ -30,6 +30,33 @@ proc run_boot_dir {binaries} {
...
+        set fh [open "[run_dir]/boot/grub/grub.cfg" "WRONLY CREAT TRUNC"]
+        puts $fh "set timeout=0"
+        puts $fh "menuentry 'Genode on Linux' {"
+        puts $fh " insmod linux"
+        puts $fh " linux /vmlinuz console=ttyS0,115200 amd_iommu=off intel_iommu=off"
+        puts $fh " initrd /initrd"
+        puts $fh "}"
+        close $fh
+
+        run_image
```
chelmuth: These statements should be run conditionally for the base-linux-hw case only, or do I miss something?

tool/run/qemu.inc
```diff
@@ -21,7 +21,6 @@ proc qemu_args { } {
 # XXX should by removed in favor of [have_include "exec/qemu"]
 #
 proc is_qemu_available { } {
-	if {[have_spec linux]} { return false }
```
chelmuth: For base-linux-*traditional* this line should be kept, i.e., base-linux-hw may fall through.

tool/run/boot_dir/linux
```diff
@@ -31,32 +31,35 @@ proc run_boot_dir {binaries} {
...
+            puts $fh "set timeout=0"
+            puts $fh "menuentry 'Genode on Linux' {"
+            puts $fh " insmod linux"
+            puts $fh " linux /vmlinuz console=ttyS0,115200 amd_iommu=off intel_iommu=off"
+            puts $fh " initrd /initrd"
+            puts $fh "}"
+            close $fh
+
+            run_image
+        }
```
chelmuth: Oh, this addresses my first remark. Please merge this into a1a765d298d686402e0114084be95d9416d13f1e

repos/base-linux/src/core/include/io_mem_session_component.h
```diff
@@ -26,6 +26,16 @@ namespace Genode {
 
 	class Io_mem_session_component : public Rpc_object<Io_mem_session>
 	{
+		private:
```
chelmuth: coding style: missing newline

repos/base-linux/src/core/include/irq_object.h
```diff
@@ -0,0 +1,38 @@
+
```
chelmuth: coding style: missing file header

repos/base-linux/src/core/spec/linux/dataspace_component.cc
```diff
@@ -0,0 +1,77 @@
+/*
+ * \brief  Linux-specific core implementation of the dataspace component
+ * \author Stefan Kalkowski
```
chelmuth: Is this the real author?

repos/base-linux/src/core/spec/linux/io_mem_session_component.cc
```diff
@@ -0,0 +1,58 @@
+/*
+ * \brief  Linux-specific IO_MEM service
+ * \author Christian Helmuth
```
chelmuth: Nope, I did not create this file ;-)

repos/base-linux/src/core/spec/linux/io_port_session_component.cc
```diff
@@ -0,0 +1,15 @@
+
```
chelmuth: coding style: missing file header

repos/base-linux/src/core/spec/pc/io_port_session_component.cc
```diff
@@ -0,0 +1,17 @@
+
```
chelmuth: coding style: missing file header

chelmuth:
Sorry for the unintended approval - the review tool confused me as I was reviewing commit by commit and just intended to approve the last commit...

repos/base-linux/src/core/spec/linux/dataspace_component.cc
jklmnn: Yes I think so. The file has originally been copied from `repos/base-linux/src/core/dataspace_component.cc`. Since `repos/base-linux/src/core/spec/pc/dataspace_component.cc` is also (an adapted) copy of this file git decided that this is the new file.

repos/base-linux/src/core/spec/linux/io_mem_session_component.cc
jklmnn: Iirc this is also the case for `repos/base-linux/src/core/spec/pc/io_mem_session_component.cc` ;)

tool/run/qemu.inc
jklmnn: How would I check for base-linux-hw here? It is currently selected by choosing KERNEL=linux and BOARD=pc.

tool/run/image/uefi
jklmnn: Initially I had the problem that the size of the generated image was too small for some cases. This is merely a quick fix. I cannot tell exactly what needs to be done to fix this properly.

jklmnn:
Thanks for the review. I addressed the code style and headers in https://github.com/genodelabs/genode/pull/2829/commits/7f412045b1ba79ec2f87acdf8f16c3bc86ec48e2 and merged the two tool commits. On the other issues I'm not completely sure how to fix them properly.

tool/run/image/uefi
chelmuth: In this case I'd drop this change or replace it by https://github.com/genodelabs/genode/blob/master/tool/run/image/disk#L28-L37

tool/run/qemu.inc
chelmuth: untested sketch: `[expr [have_spec linux] && [board] == "linux"]` ?

tool/run/image/uefi
jklmnn: This sounds fine to me.

tool/run/qemu.inc
jklmnn: I fixed that in https://github.com/genodelabs/genode/pull/2829/commits/9d9f7e365bd801a4c73c0b883f8c81c3d0bf2d29.

tool/run/image/uefi
jklmnn: I added https://github.com/genodelabs/genode/pull/2829/commits/a754be14032862c7d99df2e8bc27a3e2393ced92. I removed the hard coded disk size change and used the `-L` flag for `dm` as it is also done in the file you referenced. This seems to include the files in `var/run/<runfile>/genode` which are mostly symbolic links. It seems that these haven't been into consideration which fits with my observation that the problem appeared more often the more components the run files used.

jklmnn:
I hope I properly addressed both todos. About squashing the commits I'd like to keep the actual code and the tooling commits separated. But you could squash https://github.com/genodelabs/genode/pull/2829/commits/c7d807b07d869874b73d554fb758eaeb3615f2b8, https://github.com/genodelabs/genode/pull/2829/commits/0b8dc1e3c5b8f291d9d76f46d9949d19d5765396 and https://github.com/genodelabs/genode/pull/2829/commits/7f412045b1ba79ec2f87acdf8f16c3bc86ec48e2 into a single commit and the rest into another single one.

jklmnn:
I also added https://github.com/genodelabs/genode/pull/2829/commits/1ace6f0d47c635955c2391fad4dbb97b1c746db2 which removes an entry from `tool/boot/README` that is outdated.

tool/run/image/uefi
chelmuth: Great.

chelmuth:
The patches in this PR generate a regression with `make -C build/x86_64 run/fs_query KERNEL=linux BOARD=linux`. @jklmnn please check

jklmnn:
I've taken a look and it looks to me like that the depot package isn't up to date. The file paths (e.g. `src/core/dataspace_component.cc`) aren't up to date at least. I assume that the other errors come from incorrect include paths, etc.
I've taken a look into the base-linux recipes but to be honest I didn't really understand where I would need to change things.

For completeness, the error I'm getting:
```
In file included from /.../genode/depot/genodelabs/src/base-linux/2019-12-03/src/include/linux_dataspace/linux_dataspace.h:17,
                 from /.../genode/depot/genodelabs/src/base-linux/2019-12-03/src/core/include/dataspace_component.h:22,
                 from /.../genode/depot/genodelabs/src/base-linux/2019-12-03/src/core/dataspace_component.cc:15:
/.../genode/depot/genodelabs/api/base/2019-08-27/include/dataspace/dataspace.h:24:1: error: expected class-name before ‘{’ token
 {
 ^
    COMPILE  thread.o
    COMPILE  thread_linux.o
In file included from /.../genode/depot/genodelabs/src/base-linux/2019-12-03/src/core/dataspace_component.cc:16:
/.../genode/depot/genodelabs/src/base-linux/2019-12-03/src/core/include/region_map_component.h:70:22: error: ‘Platform_pd’ has not been declared
   void address_space(Platform_pd *) { }
                      ^~~~~~~~~~~
/.../genode/depot/genodelabs/src/base-linux/2019-12-03/src/core/dataspace_component.cc:21:39: error: variable or field ‘attached_to’ declared void
 void Dataspace_component::attached_to(Rm_region &region)
                                       ^~~~~~~~~
/.../genode/depot/genodelabs/src/base-linux/2019-12-03/src/core/dataspace_component.cc:21:39: error: ‘Rm_region’ was not declared in this scope
/.../genode/depot/genodelabs/src/base-linux/2019-12-03/src/core/dataspace_component.cc:21:50: error: ‘region’ was not declared in this scope
 void Dataspace_component::attached_to(Rm_region &region)
                                                  ^~~~~~
/.../genode/depot/genodelabs/src/base-linux/2019-12-03/src/core/dataspace_component.cc:21:50: note: suggested alternative: ‘union’
 void Dataspace_component::attached_to(Rm_region &region)
                                                  ^~~~~~
                                                  union
/.../genode/depot/genodelabs/src/base-linux/2019-12-03/src/core/dataspace_component.cc:28:41: error: variable or field ‘detached_from’ declared void
 void Dataspace_component::detached_from(Rm_region &region)
                                         ^~~~~~~~~
/.../genode/depot/genodelabs/src/base-linux/2019-12-03/src/core/dataspace_component.cc:28:41: error: ‘Rm_region’ was not declared in this scope
/.../genode/depot/genodelabs/src/base-linux/2019-12-03/src/core/dataspace_component.cc:28:52: error: ‘region’ was not declared in this scope
 void Dataspace_component::detached_from(Rm_region &region)
                                                    ^~~~~~
/.../genode/depot/genodelabs/src/base-linux/2019-12-03/src/core/dataspace_component.cc:28:52: note: suggested alternative: ‘union’
 void Dataspace_component::detached_from(Rm_region &region)
                                                    ^~~~~~
                                                    union
/.../genode/depot/genodelabs/src/base-linux/2019-12-03/src/core/dataspace_component.cc:34:6: error: redefinition of ‘void Genode::Dataspace_component::detach_from_rm_sessions()’
 void Dataspace_component::detach_from_rm_sessions()
      ^~~~~~~~~~~~~~~~~~~
In file included from /.../genode/depot/genodelabs/src/base-linux/2019-12-03/src/core/dataspace_component.cc:15:
/.../genode/depot/genodelabs/src/base-linux/2019-12-03/src/core/include/dataspace_component.h:107:9: note: ‘void Genode::Dataspace_component::detach_from_rm_sessions()’ previously defined here
    void detach_from_rm_sessions() { }
         ^~~~~~~~~~~~~~~~~~~~~~~
/.../genode/depot/genodelabs/src/base-linux/2019-12-03/src/core/dataspace_component.cc:53:43: error: definition of implicitly-declared ‘virtual Genode::Dataspace_component::~Dataspace_component()’
 Dataspace_component::~Dataspace_component()
                                           ^
    COMPILE  thread_myself.o
make[6]: *** [/.../genode/repos/base/mk/generic.mk:57: dataspace_component.o] Error 1
```

chelmuth:
@jklmnn I already fixed the depot-recipe issue on staging. What I was referring to in my comment is the follow-up error. Please see 737c1b3 for what I discovered: It is supported by the original code that `_local()` returns a `nullptr`, which is then mapped to an invalid capability.

jklmnn:
@chelmuth ah, thank you, I will check that.

chelmuth:
Please note ce3bbc0dd938b2d2aaabc1fe8c84a4eab39b3bf2, which completes this issue for me.

jklmnn:
Sorry for my late response. The commit looks fine to me, thanks for fixing it.

PR #2867, opened 2018-06-13
Python3 support

ValiValpas:
I'm not sure whether this is interesting / worth merging. Alternatively, this may go into genode-world.

Unfortunately, the run file requires some manual preparation: One must zip the `Lib` of the Python source files and make it available to the run file. The interpreter requires the hashlib module on startup so that the python libraries must always be available. Do you have any idea, how to automate the zipping of `contrib/python3-<hash>/src/lib/python3/Lib` in the run file? I don't know how to query the hash from within the run script.


nfeske:
Wow! That is very cool and unexpected.

I think that genode-world would be a better fit though as we want to limit the scope of the genode main repository to software that is either essential for Genode, or that requires a significant and ongoing integration effort (like Qt). On that account, we plan to move libsdl, dosbox, etc. to genode-world also. 

Would you like to open a corresponding pull request at genode world? I could then take the opportunity to remove python2 from genode's main repository.

ValiValpas:
Thanks, that was my thinking as well. I created the pull request here anyway because the port requires the `clock_getres` symbol from libc (see commit 07c6c055fe0344679136964b2aad0b0221de5b25). I'll open a pull request for the other commit at genode-world, though.

nfeske:
Your little libc patch is already on genode/staging and will hit the master branch soon.

PR #2882, opened 2018-06-25
Ada exception support

jklmnn:
This PR adds support to handle Ada exceptions in Genode. Furthermore it adds 64bit integer arithmetics support to the runtime. It also fixes Ada include paths by omitting the current directory.


jklmnn:
I will fix that, excuse my laziness when it comes to file headers.

jklmnn:
I added headers to all major files and went with the beautify tool over it.

repos/libports/src/lib/ada/runtime/a-except.adb
```diff
@@ -1,3 +1,13 @@
+--
+--  \brief  Ada exceptions
+--  \author Johannes Kliemann
+--  \date   2018-06-25
+--
+--  Copyright (C) 2018 Componolit GmbH
+--  This file is part of the Genode OS framework, which is distributed
```
nfeske: missing newline below copyright line

repos/libports/src/lib/ada/runtime/a-except.ads
```diff
@@ -1,3 +1,13 @@
+--
+--  \brief  Ada exceptions
+--  \author Johannes Kliemann
+--  \date   2018-06-25
+--
+--  Copyright (C) 2018 Componolit GmbH
+--  This file is part of the Genode OS framework, which is distributed
```
nfeske: missing newline

repos/libports/src/lib/ada/runtime/a-except.adb
nfeske: I wonder, would you mind also adding Genode Labs as copyright holder since the changes are contributed under the GCA. I want to make it clear that Genode Labs is in the legal position to license Genode - including these files - commercially. @senier What do you think?

senier: No objections. There is [at least one](https://github.com/genodelabs/genode/blob/master/repos/libports/src/lib/pthread/rwlock.cc) other instance where we did not do this. We can also update those other occurences with a Genode copyright header if that helps clarifying things to your customers.

nfeske:
Just perfect! Thanks for the fixups, which I promptly merged to staging.

nfeske:
On x86_32, the attempt to bulid test/ada produces the following error:

```
make[2]: *** No rule to make target '/plain/no/genode.git/repos/base/src/test/ada/startup.cc', needed by 'startup.o'.  Stop.
```

When buliding test/ada_exception on any architecture with `-O0`, the following error occurs:

```
ada.lib.so: undefined reference to `system__concat_2__str_concat_2'
```

You can enable `-O0` by adding the following line in your etc/build.conf:

```
CC_OLEVEL = -O0
```

@jklmnn May you have a look at these (hopefully minor) bulid issues?

jklmnn:
Unfortunately I cannot reproduce the error in any way. All architectures build fine, even with `-O0`. Can you provide me the exact commit ID you're building on (I assume this PR but I want to make sure) and your etc/build.conf?

jklmnn:
Okay, with `CC_OLEVEL` in etc/tools.conf I can reproduce the missing symbol. It is probably caused by a symbol being inlined on higher optimizations (or rather being not inlined on `-O0`).

But I still can't reproduce the first problem. It generally seems a bit weird to me since it looks as make somehow doesn't find startup.cc but it also seems to be architecture dependent. I have no idea what could be the cause of this. Can you give me any further details on that? Maybe any components that could be built before it and could influence this?

jklmnn:
https://github.com/genodelabs/genode/pull/2882/commits/bce7d0a5a6a4297bbb3a136ec04953348d2fc65d should fix the missing symbol.

nfeske:
Thanks for the fixup, which I merged to staging.

I can't reproduce the other problem either right now. Let us wait for the build results of the next night.

chelmuth:
With current staging building fails with
```
genode-arm-gnatmake: ".../repos/libports/src/test/ada_secondary_stack/stack.adb" compilation error
```
Does anybody know why there are no diagnostics?

chelmuth:
Thanks for the fix. Also, I can't reproduce the missing diagnostic message today, so I must have missed the line yesterday.

nfeske:
The commit series entered the master branch. Thanks @jklmnn!

PR #3044, opened 2018-11-15
ada: replace local runtime implementation with port

jklmnn:
This replaces the implementation of the Ada runtime currently residing in Genode with a port located at https://github.com/Componolit/ada-runtime. This is done to use the implementation on other platforms too without the drawback of maintaining two separate versions.

I tried to adapt my changes to the current staging but I'm not sure how these influence the depot recipes that are already available for the runtime.


nfeske:
Technically, that's a welcome change that makes the integration of the Ada runtime with Genode much more obvious.

As a non-technical implication, however, the use of the AGPLv3 as license for the runtime locks out the use of Ada for Genode components or libraries that we might wish to commercially license to 3rd parties in the future. Is this intended?

I'm just asking out of curiosity. Right now, we don't have any immediate plans. But should we go forward with our longer-term ambition of moving selected parts of the Genode's base libraries to SPARK, the license of the runtime would certainly block us.

jklmnn:
Well we have licensed the original repo with AGPL, but I honestly didn't thought of the implications of this. I think you should discuss this with @senier as he's more involved with how we should license this.

nfeske:
@jklmnn I wonder, should we better wait until the license uncertainty is cleared? Or would you prefer us to merge your pull request right away? (the code removed by your pull request is your's after all)

senier:
Using AGPL was not really a deliberate choice here, but rather is this our default license. I get your point regarding commercial customers, but I don't want to make a hasty decision on licensing here. Hence, I suggest to merge the changes and allow open source users of Genode to benefit from the runtime. We should able to come up with a suitable license that supports the commercial case until SPARK reaches significance.

jklmnn:
I hope this resolves the licensing uncertainties for now so I would prefer to merge the changes.

nfeske:
I merged the commit to staging.

@jklmnn If I understand @senier correctly, the license uncertainty remains indefinitely unresolved though. In principle, it could be avoided by adopting the license of GCC's runtime (GPLv3+linking exception) that your runtime is based on. That would lift the restriction to use GPLv3/AGPLv3 for all Ada-based Genode components. Should @senier deliberately want to keep the restriction, that's fair enough. But it creates friction.

chelmuth:
This merge breaks the ada source archive. Please fix `./tool/depot/create genodelabs/src/ada`.

m-stein:
I'll have a look at it.

m-stein:
@chelmuth I could fix the recipes:
Fixup "ada: replace local runtime implementation with port" (fix recipes)

Unfortunately, the depot_autopilot doesn't run again yet because there's still a problem with pkg/test-ldso.

m-stein:
...fixed the commit hash in the initial message.

nfeske:
Thanks @m-stein. I merged your fixup to staging.

jklmnn:
@nfeske we added the linking exception to the Ada runtime. I updated the port in https://github.com/genodelabs/genode/pull/3044/commits/a1b75eb50991ba7b53e12591e7b13cb338689126.

nfeske:
@jklmnn and @senier, thank you for addressing my concern about the runtime license. I very much appreciate your response.

Since the original commit is still on staging, would you be fine with me squashing https://github.com/genodelabs/genode/commit/a1b75eb50991ba7b53e12591e7b13cb338689126 as a fixup to the original one?

jklmnn:
I'm fine with squashing it.

nfeske:
Alright, merged to staging: https://github.com/genodelabs/genode/commit/798dde128fb45b7c4ac2b24c4b55701c21910381

PR #3053, opened 2018-11-20
zynq network improvements and board defs

ValiValpas:
There are major performance improvements for the zynq Nic driver. First of all, there is now a zero-copy implementation that directly hands over of the Packet stream buffers to the DMA. For this to work, the Packet_stream must provide the physical address of the buffers. Second, checksum offloading has been enabled. The driver also sends and receives pause frames.

In addition to this, I suggest to align the bulk buffer of the Packet stream so that it starts at the beginning of a cache line. This is not required for the zero-copy implementation though.

There is also a minor change in the defs/zynq.h: As the RAM size is board related, it should rather be specified in defs/zynq_qemu.h.

### DIFF
--- a/repos/base/include/drivers/defs/zynq.h
+++ b/repos/base/include/drivers/defs/zynq.h
@@ -30,7 +30,6 @@ namespace Zynq {
 
 		/* normal RAM */
 		RAM_0_BASE = 0x00000000,
-		RAM_0_SIZE = 0x40000000, /* 1GiB */
 
 		/* AXI */
 		AXI_0_MMIO_BASE = 0x40000000, /* PL AXI Slave port #0 */
--- a/repos/base/include/drivers/defs/zynq_qemu.h
+++ b/repos/base/include/drivers/defs/zynq_qemu.h
@@ -21,6 +21,8 @@ namespace Zynq_qemu {
 	using namespace Zynq;
 
 	enum {
+		RAM_0_SIZE = 0x40000000, /* 1GiB */
+
 		CORTEX_A9_PRIVATE_TIMER_CLK = 100000000,
 		CORTEX_A9_PRIVATE_TIMER_DIV = 100,
 	};
--- a/repos/os/include/os/packet_stream.h
+++ b/repos/os/include/os/packet_stream.h
@@ -463,6 +463,7 @@ class Genode::Packet_stream_base
 		Genode::Region_map          &_rm;
 		Genode::Dataspace_capability _ds_cap;
 		void                        *_ds_local_base;
+		void                        *_ds_phys_base;
 
 		Genode::off_t  _submit_queue_offset;
 		Genode::off_t  _ack_queue_offset;
@@ -485,6 +486,7 @@ class Genode::Packet_stream_base
 
 			/* map dataspace locally */
 			_ds_local_base(rm.attach(_ds_cap)),
+			_ds_phys_base((void*)Dataspace_client(_ds_cap).phys_addr()),
 			_submit_queue_offset(0),
 			_ack_queue_offset(_submit_queue_offset + submit_queue_size),
 			_bulk_buffer_offset(_ack_queue_offset + ack_queue_size)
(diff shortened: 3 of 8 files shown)


nfeske:
The zero-copy approach is very nice! But tainting the packet-stream API with the notion of physical addresses will potentially stand in the way in the future. Let me try to explain:

We ultimately want to remove the general ability of requesting the physical address of arbitary dataspaces in the future because the allocation of physical addresses may be misused by conspiring components as a covert channel. Only a few (trusted) components like the platform driver should be able to request that value for DMA buffers. The platform driver will, in turn, hand over this information to the respective driver. Hence, once we put this plan into practice, the `Dataspace::phys_addr` won't return a meaningful value in the general case. 

For this reason, it might be sensible to add a `packet_offset` method to the packet stream API instead of the `packet_phys` method you propose. The user of the packet stream (i.e., the driver) will always have a means to obtain the DMA address of the DMA buffer allocated from the platform driver (right now, the return value of `Dataspace::phys_addr()`). So it can manually add the packet offset to the DMA base address when programming the device.

Does this make any sense to you @ValiValpas? If yes, would you consider adjusting the pull request in this respect?

ValiValpas:
@nfeske, thanks for your feedback. I was surprised that the dataspace made the physical address available anyway, thus I get your point.

If I understand your plans correctly, the `Dataspace::phys_addr` will at some point be replaced by a means to obtain this address from the platform driver. Hence, I should move the `phys_addr()` call from the Packet_stream API to the nic_drv. Sounds reasonable from my point of view. However, I don't think any `packet_offset` method is required as the offset is already part of the Packet_descriptor or am I missing something?

nfeske:
> I don't think any packet_offset method is required as the offset is already part of the Packet_descriptor or am I missing something?

You are completely right. ;-)

ValiValpas:
@nfeske, I added two fixup commits whereby one of those is a revert of the Packet_stream API modifications. Could you squash the commits?

nfeske:
@ValiValpas great! I merged the squashed commit to staging now.

The alignment improvement sounds good too. But I wonder, may this change affect the buffer-size calculation in certain corner cases? The payload buffer is now moved a few bytes. May this movement possibly exceed the bounds of the dataspace?

I noticed the "zynq: move RAM size to board defs" commit to staging, hoping that this was your intention.

ValiValpas:
Thanks, perfect.

Regarding the alignment: The `_bulk_buffer_size` is calculated from the dataspace size and the (now aligned) offset. As the packet allocator is initialised with the correct offset and size values, I can dispel your concerns.

nfeske:
Thank you @ValiValpas for your convincing statement. ;-) Merged to staging.

ehmry:
I like this move to cache aligned packets, but I suggest that the alignment be applied at the source side of the packet stream and the sink side recognize the original unaligned and larger buffer region. Otherwise a stream sink compiled against the new header may treat packets from a stream source compiled against the old header as invalid. This is because the first packet allocated by the source will probably be at a buffer start address that is slightly lower than an aligned buffer. Patch to fix this at bf7a947eff3e6956a8b51ff6629f6928f1f9b8f7.

This raises another issue, the return value of `packet_content` is almost never checked, and is a nullptr in the case of a packet that is treated as out-of-bounds. I have a patch at 7d28fa86605f35113c8336a07b455544ce6876d8 to throw exceptions for out-of-bounds packets. Perhaps `apply_source` and `apply_sink` methods could be added to the streams so that a packet content functor is only invoked if the packet is valid, and no exceptions would need to be caught?

nfeske:
As discussed offline, I'd prefer keeping @ValiValpas version of the alignment patch, for two reasons. First, the implementation is more obvious to the reader. Intuitively, sink and source should follow exactly the same convention. Second, the problem solved by @ehmry's commit is merely one of an ABI incompatibility. So it is just temporary until all pkgs are updated to the new version of Genode.

The latter problem observed by @ehmry - the apparently missing packet-validity checks in several components - is important to fix. Throwing a meaningful exception (instead of returning a nullptr) is a good first step.

nfeske:
@ValiValpas I wonder, would you like to write a few words about your Zynq work for the release notes of Genode 18.11?

ValiValpas:
@nfeske sure.

nfeske:
@ValiValpas We plan to publish the release today and are currently reviewing the release notes. If you send me the text snippet before 12:00, I'll be happy to integrate it. Otherwise, I'll try to find some words.

ValiValpas:
@nfeske I just created a pull request on your release_notes branch.

nfeske:
Merged into master.

PR #3225, opened 2019-03-11
Support Ada.Real_Time and exit status in SPARK runtime

senier:
This is an update to Genode's Ada runtime. Changes:

# SPARK components

Components using the `spark` runtime can be realized easier and safer now by using `spark/component.h`. A component implements `Spark::Component::construct()` which either hosts an Ada main program that is finished on return or a setup routine of a component that operates in an event-based fashion. As Ada requires finalization code to run on component exit, we distinguish these to scenarios by the return value of `construct()`:

```c++
enum Result {
  EXIT,   /* Run finalizatoin code an exit component */
  CONT   /* Continue execution after construct() */
};
```

**We do not provide an automated way of running finalization code for event-based components, yet.**

# Exit status

The exit status of SPARK components can now signaled to the parent component using the standard `Ada.Command_Line` interface when using `spark/component.h`:

```ada
with Ada.Command_Line;

...
Ada.Command_Line.Set_Exit_Status (42);
...
```

**This only works for components that implement their main program inside `construct()` and return `Spark::Component::EXIT`.**

# Ada.Real_Time

The standard package `Ada.Real_Time` can now be used on Genode to perform time measurements:

```ada
with Ada.Real_Time;

procedure Real_Time
is
   Start  : Ada.Real_Time.Time;
   Diff   : Ada.Real_Time.Time_Span;
begin
   Start := Clock;
   -- Do_Something
   Diff := Clock - Start;
   --  ...
end Real_Time.
```

To use `Ada.Real_Time`, the component needs a `Timer` session.

# Genode session integration

No pointers to session instances need to be passed to the runtime anymore. The only requirement is a pointer to a `Genode::Env` instance:

```c++
extern Genode::Env *__genode_env;
```

Sessions (`Timer`, `Terminal`) are constructed by the runtime if and when they are required. All this is automatic when using `spark/component.h`.

# Misc

Silent those annoying `"Get_Jmpbuf_Address_Soft not implemented"` warnings.


m-stein:
This is really cool! I already tried it out successfully and I'm looking forward to use it more. However, it seems that the commits do not bring the required updates for the spark recipes. I wanted to translate the spark_real_time test to a test package for the depot autopilot and started modifying recipes/api/spark accordingly but I got stuck at the point where we need an a-reatim.ali for the a-reatim.ads that I had to introduce (it seems you have avoided the use of the spec file but I think for the API archive its necessary). How can I achieve to create a proper ALI file?

senier:
> This is really cool! I already tried it out successfully and I'm looking forward to use it more. However, it seems that the commits do not bring the required updates for the spark recipes.

Thanks for trying! Indeed, the missing runtime .ali files are an oversight.

> I wanted to translate the spark_real_time test to a test package for the depot autopilot and started modifying recipes/api/spark accordingly but I got stuck at the point where we need an a-reatim.ali for the a-reatim.ads that I had to introduce (it seems you have avoided the use of the spec file but I think for the API archive its necessary).

For run files, the specs are imported from the port via `INC_DIR`. For the API archive, you have to add them.

> How can I achieve to create a proper ALI file?

The process is as follows:

1. Build the Ada sources and find the created ALIs
2. Remove compiler switches (lines starting with 'A') as their are potentially architecture dependent
3. Replace timestamps of the form `20190308061037` by all zeros (`00000000000000`)
4. Fork [our ALI repo](https://github.com/Componolit/ada-runtime-alis)
5. Add the new file under the `alis` directory
6. Create a pull request
7. Add the new ALI to `MIRROR_FROM_ADA_ALI_DIR` in `repos/libports/recipes/api/spark/content.mk`
8. Update `repos/libports/ports/ada-runtime.port`

m-stein:
@senier Thanks a lot for this detailed description!

I've created a fixup:
ca7f199e83 Fixup "Support Ada.Real_Time and exit status in SPARK runtime" (test via depot_autopilot)

Plus the required patch for the ALI's:
https://github.com/m-stein/ada-runtime-alis/commit/2e236fc8b1a8b94ba0963bef6d2cd78301f97615

senier:
@m-stein Thanks! I merged your commit into master of ada-runtime-ali and also updated the ports file.

m-stein:
@senier Ah, forgot it, thanks for updating the port files ;)

nfeske:
I merged the three commits to staging now.

nfeske:
The  test-spark_realtime test still fails on the following platform:

* linux_x86_32
* hw_imx6q_sabrelite
* sel4_imx6q_sabrelite
* foc_imx6q_sabrelite
* sel4_x86_32 (hardware)
* pistachio_x86_32 (hardware)
* okl4_x86_32 (hardware)
* nova_x86_32 (hardware)
* foc_x86_32 (hardware)
* fiasco_x86_32 (hardware)
* hw_rpi
* qemu/sel4_x86_32
* qemu/pistachio_x86_32
* qemu/okl4_x86_32
* qemu/nova_x86_32
* qemu/hw_zynq_qemu
* qemu/hw_pbxa9
* qemu/foc_x86_32
* qemu/foc_pbxa9
* qemu/fiasco_x86_32
* odroid_xu/hw_odroid_xu
* imx53_qsb/hw_imx53_qsb_tz
* imx53_qsb/hw_imx53_qsb
* arndale/hw_arndale
* arndale/foc_arndale

@m-stein when I merged your commit https://github.com/genodelabs/genode/commit/997ffc72995430c286859c3cb2691e655593ffd6 on Thursday, I was hopeful that you actually validated your fix to some degree. But by looking at the list above, I'm quite disappointed that you apparently haven't even tried out your fix at least on a few of the (many) platforms. It looks like just a shot in the dark. So this issue keeps blocking the next master update, even though this situation could have been avoided by manually testing a few of the platforms (at least qemu). Could you please take on this issue again?

nfeske:
I temporarily removed the commits from staging for the master update today. You can find the commits re-applied to the new staging branch. @senier would you mind executing the test on (a few of) the platforms listed above and fixing the problem? @m-stein I'd appreciate you lending @senier a helping hand.

m-stein:
@nfeske I actually tested my first commit on several platforms that are still failing.
@senier I'd be glad if I can be of help. I plan to have a closer look at it myself tomorrow again.

chelmuth:
First, I sorted the spark symbols file in 4362e40bb0ea0f3a371f189e39bcf1432a5afa92 to make it easier to track the API changes (and was astonished by the amount of changes in comparison to a sorted master symbols file).

Second, I doubt the addition of a global `__genode_env` pointer is the right direction for the ada-runtime integration as we strive for less side effects and not more. @senier do you see a chance to add an initialization function to ada-runtime, which is passed a reference/pointer to `Genode::Env` from `Component::construct()`? Also, I'd greatly appreciate the migration of the existing spark tests to the new `Spark::Component` scheme instead of the current somewhat rough approach of adding magic global symbols.

m-stein:
Just as an intermediate state info: When doing a 1sec tick 20 times in test-spark_real_time, I get the following results on pbxa9/hw:
```
[init -> depot_autopilot] 4.973 [init -> test-spark_real_time] put_us 1065661000
[init -> depot_autopilot] 6.066 [init -> test-spark_real_time] put_us 1057353000
[init -> depot_autopilot] 7.174 [init -> test-spark_real_time] put_us 18446744070498879320
[init -> depot_autopilot] 8.272 [init -> test-spark_real_time] put_us 1074437000
[init -> depot_autopilot] 9.382 [init -> test-spark_real_time] put_us 1083335000
[init -> depot_autopilot] 10.477 [init -> test-spark_real_time] put_us 1072074000
[init -> depot_autopilot] 11.586 [init -> test-spark_real_time] put_us 18446744070501865320
[init -> depot_autopilot] 12.679 [init -> test-spark_real_time] put_us 1077242000
[init -> depot_autopilot] 13.796 [init -> test-spark_real_time] put_us 1099353000
[init -> depot_autopilot] 14.899 [init -> test-spark_real_time] put_us 1078872000
[init -> depot_autopilot] 16.022 [init -> test-spark_real_time] put_us 18446744070509302320
[init -> depot_autopilot] 17.111 [init -> test-spark_real_time] put_us 1083172000
[init -> depot_autopilot] 18.189 [init -> test-spark_real_time] put_us 1073110000
[init -> depot_autopilot] 19.270 [init -> test-spark_real_time] put_us 1059316000
```

nfeske:
I propose to remove the patch series from staging until this issue is sorted out by @senier, and @chelmuth's remarks are covered. After all, I see no pressing need for the immediate integration this feature. Let's better give it some time to mature.

Removing the commits would also relieve the pressure from @m-stein to sort out the problems.

m-stein:
One problem seems to be definitely that the ``Clock`` call returns microseconds values truncated to 32 bit on 32 bit platforms and wraps are not considered by the test.

nfeske:
Please find the current commit series preserved at my spark_real_time topic branch: https://github.com/nfeske/genode/commits/spark_real_time

m-stein:
As Time/Time_Span are always 64 bit with the current runtime regardless of the maximum Clock return value we cannot handle wraps in the test without guessing the maximum Clock return value.

m-stein:
I second the opinion of @chelmuth regarding the ``__genode_env``. It is really cool to have it as quick intermediate solution for GNAT.IO, but for a longterm solution, IMO, it would be much better to let the user explicitely initialize only the required side-effect-causing modules (without using a global state).

senier:
> As Time/Time_Span are always 64 bit with the current runtime regardless of the maximum Clock return value we cannot handle wraps in the test without guessing the maximum Clock return value.

Just to make sure I got this right: The `elapsed_us()` call of the timer session returns `uint64_t`, but the value is silently capped to 2^32 on 32-bit platforms? If that is the case, the only sensible way of handling this is to disable the test and remove support for `Ada.Real_Time` on 32 bit platforms. The Ada standard requires minimum duration and precision for this interface:

> The range of Time values shall be sufficient to uniquely represent the range of real times from program start-up to 50 years later. Tick shall be no greater than 1 millisecond. Time_Unit shall be less than or equal to 20 microseconds. 
> Time_Span_First shall be no greater than –3600 seconds, and Time_Span_Last shall be no less than 3600 seconds.

I'm still not sure I understand that correctly. Is the time value in nanoseconds internally? If so, this would wrap after 4.29 seconds for a 32 bit datatype which would explain the result. If in microseconds, it should wrap only after ~70 minutes which would *not* explain the issue we're seeing.

senier:
> @senier do you see a chance to add an initialization function to ada-runtime, which is passed a reference/pointer to `Genode::Env` from `Component::construct()`?

Using an initialization function in the platform code would certainly be possible. This would also improve type safety, as we can pass a reference to `Genode::Env` and do the dirty pointer stuff internally. For some reason we failed to consider that approach, thanks for pointing out.

senier:
After an offline discussion with @nfeske we concluded that we want to follow a different approach. Initially it was tempting to use standard Ada interfaces like `Ada.Real_Time` because Ada programmers have an API they are used to. However, this introduces ambient state and the need for magically transferring Genode sessions to the Ada runtime, depending on what Ada feature is being used. We agree that making the use of sessions explicit also in the SPARK world is very desirable, such that you can be certain a component has not external effects, unless it holds a reference to a session. Also, we don't expect existing Ada code to be ported to components using the minimal SPARK runtime, such that supporting standard interfaces is not a significant benefit.

We already have a collection of [Ada interfaces](https://github.com/Componolit/ada-interface) which provides generic abstractions of platform interfaces (like logging, block devices etc.). As a result of this discussion, we'll add a timer session there, in favor of `Ada.Real_Time`. We'll also change our interfaces to include a "capability" object required for sessions to access resources on the platform. On Genode, this would be used to pass a pointer to `Genode::Env` to our sessions.

We'll also move the component startup and exit code handling into our library. For SPARK components on Genode that have a C++ main program, handling startup and exit is unnecessary. You only need to call `void adainit()` before running any Ada subprogram and handle startup and exit in your C++ code.

The only improvement left is muting the `Get_Jmpbuf_Address_Soft` warnings in the runtime. We'll upstream this in a separate PR and close this one.

m-stein:
@senier Thanks a lot for summarizing the outcome of todays discussions!

Issue #3238, opened 2019-03-19
depot_autopilot: still some sporadic false-positives


skalk:
Since the integration of the `depot_autopilot` into the nightly test routines, there are sporadic false-positive errors. As an example, tonight the `depot_autopilot` reported that the `test-magic_ring_buffer` failed due to a 10 sec timeout on imx6q_sabrelite on the Hw kernel, although it printed the good-case message immediately within the first second.
This is just an example we could observe these failing of actually succeeding test regularily.

m-stein:
My last state regarding this topic is that I was not able to reproduce it. It was the reason we added the TEST_REPEAT variable to the depot_autopilot.run script. But doing tests in a loop didn't trigger it neither.

skalk:
@m-stein thanks for documenting the last approach. Nevertheless, we have to get rid of the false messages. If you cannot reproduce it, my suggestion would be to have some non-invasive interims changes on staging to identify the actual cause. It always happens on timeouts of a test. Maybe there is a way to check the output again afterwards using different mechanisms to detect the error case?

m-stein:
As a first step, we could write the log output of each test to a RAM buffer. If the test fails, we write the buffer content as hex values to to the output. This way, we could detect abnormal non-printable characters. As a second step, we could instead write the parsing steps of the Depot Autopilot to the buffer. I.e. the path taken for each character in the matching loop. Reading this is annoying work but it should definitely reveal the problem. I'll come up with appropriate patches.

skalk:
That sounds good, as long as it is printed in the error case only. Interestingly, tonight it was again `test-magic_ring_buffer` this time on top of Fiasco.OC/PBXA9 on Qemu.

m-stein:
These are some patches I'd like to suggest for staging only to be able to debug the nightly tests:
a5e461e3a7 NOT FOR MASTER "depot_autopilot.run: raise depot autopilot quota"
3b51b3e0ff NOT FOR MASTER "depot_autopilot: list log matches"
4855b19f16 NOT FOR MASTER "depot_autopilot: hex dump of failed test log"

They let the depot autopilot print some extra debug output for failed tests. First, a hex dump of the tests log is printed as can be seen in this example after the test has finished: 
```
[init -> depot_autopilot] --- Run "test-log" (max 5 sec) ---
[init -> depot_autopilot] 
[init -> depot_autopilot] 2.442 [init -> test-log] hex range:          [0e00,1680)
[...]

PR #3371, opened 2019-05-22
ada-runtime: update to GCC 8.3.0

### LINKED ISSUE #3362: ada-runtime: update to GCC 8.3.0
The current 6.3.0 version of the runtime appears to work with the upcoming Genode 19.05 tool chain with some small adjustments [1], but it might be useful to update the runtime files to version 8.3.0.

@senier, @jklmnn: would you be interested in doing this update? My issue3307 branch [2] can be used as basis for testing with the new tool chain.

[1] https://github.com/cproc/genode/commit/07078f96a0c518647532f87fe26e5a0607648983
[2] https://github.com/cproc/genode/tree/issue3307

jklmnn:
This should fix #3362. It brings the recipes, symbols and ALI files in sync with the current version of the ada-runtime.
Changes in the ada-runtime:
 - update to GCC 8.3.0
 - removal of incomplete threading support
 - removal of incomplete image support
 - proof of secondary stack allocator
 - removal of Ada.Real_Time
 - removal of all dependencies on Genode::Env

The ALI files have been updated in the port repository so the patch could be removed, too.


chelmuth:
The commit triggers an error in *test-spark_exception* of the depot autopilot tests.
```
[2019-05-23 08:37:02] [init -> depot_autopilot] --- Run "test-spark_exception" (max 20 sec) ---
[2019-05-23 08:37:02] [init -> depot_autopilot] 
[2019-05-23 08:37:02] [init -> depot_autopilot] 0.103 [init -> test-spark_exception] Ada exception test
[2019-05-23 08:37:02] [init -> depot_autopilot] 0.104 [init -> test-spark_exception] Error: Exception raised: explicit raise in except.adb:6
[2019-05-23 08:37:02] [init -> depot_autopilot] 0.105 [init -> test-spark_exception] Error: Uncaught exception of type 'Ada::Exception::Program_Error'
[2019-05-23 08:37:02] [init -> depot_autopilot] 0.106 [init -> test-spark_exception] Warning: abort called - thread: ep
[2019-05-23 08:37:02] [init -> depot_autopilot] 0.114 [init] child "test-spark_exception" exited with exit value 1
```

chelmuth:
You may use `make -C build/x86_64 run/test PKG=test-spark_exception KERNEL=linux` to reproduce the error.

jklmnn:
Okay, the exception handling in the runtime was broken. The exception would not be propagated back to the component but would be caught by Genodes exception handler.
I fixed this and the output is equal to what the test expects.
Although even if the output is equal, the test doesn't seem to recognize this. Is there some way to debug what the test actually matches on?

jklmnn:
https://github.com/genodelabs/genode/pull/3371/commits/344e434fb1e176e93790a72d52a71edb0e73b0be should fix this. I fixed the exception propagation in the runtime. I also only match on a successful return value for the test which is sufficient to check functionality it is testing.

m-stein:
@jklmnn We have some patches for the depot autopilot that make him print out diagnostics about mismatches:
https://github.com/genodelabs/genode/issues/3238#issuecomment-475353422

These patches are not for master but they become very handy when debugging log events that do not match.

jklmnn:
Thanks for the link. For this change I'll stay with the current version as the exit code is sufficient for the test.

chelmuth:
Unfortunately, we're not there yet as `make run/test PKG=test-spark_exception KERNEL=hw BOARD=pbxa9` gets stuck.

```
kernel initialized
ROM modules:
 ROM: [700fe000,700fe679) config
 ROM: [2015d000,2015e000) core_log
 ROM: [701b4000,701e6518) init
 ROM: [7011d000,701b02ec) ld.lib.so
 ROM: [2015b000,2015c000) platform_info
 ROM: [700ff000,7010d4b4) spark.lib.so
 ROM: [701b1000,701b3684) test-spark_exception
 ROM: [7010e000,7011c9b0) timer

Genode sculpt_ce-346-g758520433d
764 MiB RAM and 64535 caps assigned to init
[init -> test -> test-spark_exception] Ada exception test
[init -> test -> test-spark_exception] Error: Exception raised: explicit raise in except.adb:6
```

jklmnn:
I took a look into it and the problem is that `__gnat_personality_v0` is called with invalid values. The first two arguments are the version and exception phase where the version is usually 1 and the phase 1 or 2. Usually we don't check that and simply return `_URC_CONTINUE_UNWIND` which indicates that we don't have an exception handler and then the Genode exception handler gets called. In your case the the values are as follows:
```
Error: __gnat_personality_v0 called with invalid values version=0 phase=867644 class=1075833200 exception=0x1009588 context=0xffffffff
```
As those values are also invalid for all Genode handlers (also note that context should be a pointer to the current exception state) all handlers simply pass them around basically resulting in a busy loop.
I have no idea where those values come from. I assume the calls to this function are generated by the compiler so this could be either caused by a broken stack or ABI incompatibilities (although this doesn't happen on x86_32). There are references on special ARM EABI personality handlings in `repos/base/src/lib/cxx/unwind.c` so ARM could have some special exception handling code but I don't know anything about that.

The only "solution" I can come up with now would be checking for this case would be indicating that the exception cannot be handled anymore which then causes Genode to catch it and kill the component. But that would break the ability to catch Ada exceptions in C++.

chelmuth:
The `class` value looks like a pointer to the stack `1075833200 == 0x401fe970`, which hints there may be an issue with the stack (maybe alignment). I have no further idea and are also lacking time to investigate before the release.

jklmnn:
I added the value check in https://github.com/genodelabs/genode/pull/3371/commits/681e50bb7dcf8e083645fcc1946e848559821d52 so there's at least a clear message when this happens. As the test still doesn't pass you might disable it on this platform (or ARM, or completely, whatever fits best). We're discussing exception support in general anyway since it introduces issues in other places, too.

PR #3465, opened 2019-07-29
util/fifo: Change empty() method of fifo to const.

### LINKED ISSUE #3464: util/fifo: change empty function to const
In repos/base/include/util/fifo.h, the empty() function should be const.

throwException:
Fixes #3464


nfeske:
Thank you. Merged to staging.

PR #3480, opened 2019-08-21
libc: Allow private writable mappings

senier:
`mmap` is restricted to read-only mappings right now. As changes in memory are not carried through to the underlying file, this is the only sound option for shared mappings. However, as I'm interpreting [POSIX](https://pubs.opengroup.org/onlinepubs/009695399/functions/mmap.html), writable private mappings should be compatible with the current semantics:

> If MAP_PRIVATE is specified, modifications to the mapped data by the calling process shall be visible only to the calling process and shall not change the underlying object. It is unspecified whether modifications to the underlying object done after the MAP_PRIVATE mapping is established are visible through the MAP_PRIVATE mapping.

Any objections to relax the constraints of `mmap` for cases where `PROT_WRITE` is used together with `MAP_PRIVATE`?


nfeske:
This looks fine to me.

PR #3514, opened 2019-10-05
Fix invalid ram session quota accounting in lan9118.

tworaz:
This driver manually checks if the RAM quota is big enough for the
Lan9118 nic session component. The problem is Root_component::_create
from which Root_component::_create_session gets called does already
check this. No need to account for it twice.


nfeske:
Thank you very much for the fix, which I just merged on our staging branch. I merely shortened the first line of the commit message to fit 50 characters (good practice) and hope that's ok.

PR #3530, opened 2019-10-15
Fix MAC address printing in dde_linux USB net drivers

OlegGirko:
Passing array of unsigned chars to `Genode::log()` function makes it converted to void pointer, resulting in printing its address.

Wrapping this array into `Genode::Cstring` solves this problem and makes it being printed properly as zero-terminaled string.

### DIFF
--- a/repos/dde_linux/src/drivers/usb/nic/nic.cc
+++ b/repos/dde_linux/src/drivers/usb/nic/nic.cc
@@ -666,7 +666,7 @@ struct sk_buff *skb_dequeue(struct sk_buff_head *list)
  ** linux/inerrupt.h **
  **********************/
 
-static void snprint_mac(u8 *buf, u8 *mac)
+static void snprint_mac(char *buf, u8 *mac)
 {
 	for (int i = 0; i < ETH_ALEN; i++)
 	{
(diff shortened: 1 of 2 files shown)


OlegGirko:
@chelmuth: I have two questions.
1. What's the reason for creating separate commit instead of merging this branch?
2. Should I create my next PR using a branch created on top of `staging` branch instead of `master`?

OlegGirko:
Actually, it will be more convenient for me to base my changes on top of `staging` branch.

I have a branch with more changes, and I'm going to make a PR from each commit in this branch. If I rebase this branch on top of `staging`, after each PR is merged, I can rebase my branch on top of `staging` branch again, and this will cause no conflicts, just less remaining commits in my branch.

Also, it will be more convenient for you: no need to `git cherry-pick` or rebase by change on `staging` branch, just merge using fast forward.

chelmuth:
> Actually, it will be more convenient for me to base my changes on top of `staging` branch.

Feel free to do so.

> Also, it will be more convenient for you: no need to `git cherry-pick` or rebase by change on `staging` branch, just merge using fast forward.

I doubt this will happen often as we merge from a couple of sources at the same time. Cherry-picking is no additional overhead for me as I have to review the commits anyway.

PR #3536, opened 2019-10-24
sntp_client: No need to limit this scenario to x86

OlegGirko:
It runs pretty well on Raspberry Pi, for example.


chelmuth:
We added this limitation to prevent our nightly autpilot runs to start the script on platforms beside x86, so I'd suggest the following alternative change.

```diff
--- a/repos/libports/run/sntp_client.run
+++ b/repos/libports/run/sntp_client.run
@@ -1,4 +1,4 @@
-assert_spec x86
+if { [get_cmd_switch --autopilot] } { assert_spec x86 }
 
 if { [have_spec linux] } {
 	puts "Run script is not supported on this platform."
```

Does this fit your needs?

OlegGirko:
As you wish. :-)

Just out of curiosity, why not test this scenario on other platforms?

chelmuth:
> Just out of curiosity, why not test this scenario on other platforms?

The reason is, we already run dozens of low-level tests (including network tests) on 24 arch-board-kernel combinations each night. The SNTP test only tests a network protocol implementation which calls for 32-bit and 64-bit compatibility but not for the entire diversity of hardware. Would you agree?

OlegGirko:
OK, if you think that this scenario is not that interesting to be tested on all arch-board-kernel combinations, I won't argue.

I just think that it would look cleaner if this was reflected in exception rules of your autopilot setup, not in the scenario itself.

OlegGirko:
But if you added a commit to `staging` branch already, let it be for now.

chelmuth:
> I just think that it would look cleaner if this was reflected in exception rules of your autopilot setup, not in the scenario itself.

You're right, it would be optimal to separate run scripts incl. whitelists of blessed platforms from our nightlies, but currently all implemented test scenarios *are* our autopilot setup and we maintain our exception rules in one place - the run scripts.

PR #3576, opened 2019-12-09
Seoul improvements

alex-ab:


repos/ports/src/app/seoul/component.cc
```diff
@@ -785,7 +785,9 @@ class Machine : public StaticReceiver<Machine>
 		Seoul::Network        *_nic          { nullptr };
 		Rtc::Session          *_rtc          { nullptr };
 
-		Vcpu *                 _vcpus[32] { nullptr };
+		enum { MAX_CPUS = 8 };
+		Vcpu *                 _vcpus[MAX_CPUS] { nullptr };
+		Genode::Bit_array<64>  _vcpus_active { };
```
chelmuth: Is this a limitation of `Bit_array` or what's the reason for the different values (8 vs. 64)?

alex-ab: Yes, the used elements in the bit array must be a multiple of 32 on 32bit and 64 on 64bit. I can change it to something like sizeof(unsigned) * 8 to save a word on 32bit builds, if you want.

chelmuth: No change needed, I was just curious (and too lazy to look it up myself). Thanks!

PR #3581, opened 2019-12-13
base-linux: enabled seccomp and socket descriptor caps for RPC

throwException:
base-linux uses seccomp to reduce the available system calls
to the minimum set needed to run base-linux. There are still
some syscalls that allow accessing global state which should
be further reduced.

base-linux uses socket descriptor to enforce capabilities,
so client cannot access RPC objects of other clients by
guessing badges any more.

the combination of seccomp and socket descriptor caps should
provide an intermediate level of security for base-linux
thereby enabling base-linux as a migration path to the
future use of microkernel-based genode.

some runscripts need more than 4096 file descriptors now,
so not only ulimit but the system limits need to be elevated.
(see /etc/security/limits.conf on debian/ubuntu systems.)


throwException:
@nfeske request for comments and the prospect of integration into future genode releases.

nfeske:
@throwException Wow! That is an impressive achievement. I greatly appreciate your pull request. It turns the base-linux variant to Genode into more than just a development vehicle. Thank you for exploring this difficult terrain.

Going forward, I'd like to split the big topic into multiple smaller steps to be merged one after another. I suggest three steps.

1. Introducing the hook for hosting the fdset as part of the `Rpc_entrypoint` into all base platforms. This is the only step that touches base platforms other than Linux.

2. Changing the IPC mechanism on base-linux to the use of select.

3. Applying seccomp to base-linux.

I'm afraid that splitting the single commit into these three concerns is quite a bit of work. Would you be ready to take it on? Otherwise - should you feel bothered by me - please let me know. ;-)

Regarding the first step (fdset). Here are my thoughts:

increasing the ABI footprint of the `Rpc_entrypoint` by 32 KBytes across all platforms is too invasive. How about the following approach?

* A new class `Rpc_entrypoint::Native_context` should be the designated container for the fdset. The class is forward-declared in the header `include/base/rpc_server.h` and defined in a framework-private header `src/include/base/internal/rpc_native_context.h`. There are two versions of this header, one for base-linux where the class holds the fdset, and an empty one used by the other platforms.

* `rpc_reply_wait`  takes a reference to an `Rpc_entrypoint::Native_context` as last argument (not a reference to `Rpc_entrypoint`). This keeps the purpose of the argument tightly under control. The argument can be used as is. No cast from an opaque array is needed.

* The `Rpc_entrypoint::Native_context` is allocated as local variable on the stack of the `Rpc_entrypoint::entry` method. This way, we avoid inflating the size of the `Rpc_entrypoint` object at the ABI level. At the constructor of the `Rpc_entrypoint`, we should add `sizeof(Native_context)` to the `stack_size`.

As a minor - formal - remark, I would very much appreciate you adding a copyright line with Genode Labs to each new file to clarify that the source file is part of the product of Genode Labs. Otherwise, commercial licensees of Genode may raise eyebrows when encountering files where Genode Labs is not mentioned as copyright holder.

I think we should discuss the steps 2 and 3 once step 1 is in shape. Do you agree?

Again, let me express my excitement about your line of work! What a cool outcome of the community summer.

throwException:
@nfeske I agree with your plan although I cannot promise when I can enact it.

As for the `Rpc_entrypoint::Native_context`: I would also like to take into account the number of capabilities needed per process as the maximum is only ever needed in `core`. I'm not yet sure, how to do this, but I think the capability quote should be an upper limit to selected/polled file descriptors.

I will also add the genode labs copyright notices wherever it's missing.

throwException:
@nfeske Through the test-init I discovered inconsistency: Because `Rpc_entrypoint::_alloc_rpc_cap` in my implementation creates new caps via `lx_socketpair` without an RPC to core or parent, the accounting of caps doesn't work as expected.

I'm not quite sure how to fix this in a consistent and useful way. Do you have an idea?

nfeske:
To uphold the consistency, I'd recommend performing a call to core's `alloc_rpc_cap` in addition to the local creation of the socket pair. The linux-specific implementation at core would merely adjust the accounting.

In the meantime, I came up with an idea how to allocate the `Native_context` such that the size can differ between core and non-core as you requested. How about adding a new `Rpc_entrypoint::_entry(Native_context &)` method that executes the code we currently have in `Rpc_entrypoint::entry()`? On base-linux, the `entry()` would then look like this:

```
void Rpc_entrypoint::entry() override
{
  Native_context<1024> native_context { };
  _entry(native_context);
}
```

Here, `entry` is just a wrapper around `_entry` for the sole job of allocating the native context. The important detail is that `native_context` is allocated on the stack. The `entry` method could now be implemented differently for core and non-core, and differently between kernels. The actual `_entry` method can remain generic.

Note that we will also need to consider the size of `native_context` when dimensioning the stack size in the `Rpc_entrypoint` constructor. I would introduce a new method `Rpc_entrypoint::_native_stack_size(size_t)` method that filters the `stack_size` value when passed to the `Thread` constructor. The implementation would sit in the same compilation unit as `Rpc_entrypoint::entry` and roughly look as follows (just a sketch):

```
size_t Rpc_entrypoint::_native_stack_size(size_t size) const
{
  return size + sizeof(Native_context);
}
```

Does that sound feasible?

throwException:
> To uphold the consistency, I'd recommend performing a call to core's `alloc_rpc_cap` in addition to the local creation of the socket pair. The linux-specific implementation at core would merely adjust the accounting.

I've implementet that and it seems to work nicely.

> Does that sound feasible?

@nfeske I've hit a problem with the implementation of placing the native context on the stack. I get a segfault despite having increased the stack size as you suggested. Is there anything more I should take care of when storing ca. 128 KiB on the stack.

nfeske:
I am afraid that your problem description is too vague. Please consider the following to investigate:

- Take a look at the page-fault address and page-fault IP as revealed by the `dmesg` command. Is the page-fault address a stack address? Inspect the code at the IP via `objdump -lSd debug/core-linux`.
- Find out which component is dying. Is it core? I'd guess so.
- Add a call to `wait_for_continue` early on during the startup of the crashing component. When the scenario reaches this point, it waits for the press of the enter key. At this point, attach gdb to the process via `gdb -p`. Please refer to http://genodians.org/ssumpf/2019-04-26-java-gdb for more details.
- Instrument the `Thread::_alloc_stack` method. Does it receive the the correct stack size?

throwException:
@nfeske I did the following:

1. Found and fixed the segfault, so the `Native_context` is now on the stack

2. Changed from `poll` to `epoll` for better performance. This also reduces the need to use much stack memory for the `Fdset` as this is now managed by the kernel.

3. Split the merge request into 3 commits as you suggested.

I think it should now (almost) be ready for integration.

nfeske:
Thanks for sharing these great news.

While reviewing the first commit, I wonder, could the argument that you added to `ipc_call`, `ipc_reply_wait`, and the `Ipc_server::Ipc_server` be a `Native_context`. This would clarify the very narrow potential side effects. In contrast, by passing a `Rpc_entrypoint &` as done by your current patch, the functions/methods have a much larger (but unneeded) potential reach.

Please let us move the `Native_context` into the scope of the `Rpc_entrypoint`. It is enough to just forward-delare it in the public header. I'm asking for this change because the term "context" is too general to be meaningful without context. ;-) E.g., there already exists a signal context, there is also the notion of a context being a thread. So let us better be clear about it.

The `Native_context` does not need to have a vtable. I guess you added the virtual destructor in order to safely inherit the specific class from the one defined in `rpc_server.h`. But I would not define it in the public header at all. Just forward-delare it, and leave the definition local to the implementation of the base library and core.

There is a whitespace mistake (tabs/spaces) in the comment of `_entry`. When scrolling through the patch in GitHub you can spot a few other glitches like this.

Please try to avoid the use of pointers, and better use references, or better avoid both whenever possible. In my understanding, there is no need to store/pass `Native_context` as pointer at all.

Please put the `const` keyword after the type. I.e., instead of
```
Rpc_entrypoint(const Genode::Rpc_entrypoint&)
```
better write
```
Rpc_entrypoint(Genode::Rpc_entrypoint const &)
```
This is a matter of coding style. We adhere to it in order to be consistent with the general rule in C++ that `const` refers to the thing on its left.

Instead of manually deleting the assignment operator / copy constructor, you may better inherit the `Noncopyable` utility.

Please do not extend the public `Rpc_entrypoint` interface. The added accessor to obtain a pointer to the `Native_context` is no-go. I think we can live without it.

Would you like to adjust the patch? If I'm bothering you with my nitpicking, please tell me. So I can step in.

throwException:
> Would you like to adjust the patch?

Yes, I will adjust the patch. We also discovered a few things in our internal review.

throwException:
@nfeske

> Please try to avoid the use of pointers, and better use references, or better avoid both whenever possible. In my understanding, there is no need to store/pass `Native_context` as pointer at all.

We need to keep the `Native_context* _native_context` in `Rpc_entrypoint` as it isn't initialized at construction, but only at `_entry`.

> Instead of manually deleting the assignment operator / copy constructor, you may better inherit the `Noncopyable` utility.

I didn't manage to use the `Noncopyable` as it's already inherited indirectly via `Object_pool`.

The other points and a few things from our review are improved.

nfeske:
The new commit looks very nice already! Thanks for the adjustments.

> We need to keep the Native_context* _native_context in Rpc_entrypoint as it isn't initialized at construction, but only at _entry.

Could you please point me to the place where this pointer is de-referenced? I somehow fail so see it and assumed that the native context never leaves the evades the scope of the `entry` method.

Even though the patch is already in a very good shape, I still have a few remaining remarks:

* `_native_stack_size` should be a private class function of `Rpc_entrypoint` (`static size_t _native_stack_size(size_t)`)

* Please leave the two-lines vertical space between class definitions in tact. The same goes for method implementations in cc files. See https://genode.org/documentation/developer-resources/coding_style#Vertical_whitespaces for the conventions.

* There is no need to fully qualify `Rpc_entrypoint::Native_context` within the scope of `Rpc_entrypoint`. I.e., the signature of `_entry` should take `Native_context &` as argument, not `Rpc_entrypoint::Native_context`.

* Please indent the `private` and `public` sections of the `Ipc_server` definition according to our coding style. Please pay attention to the vertical whitespace as well. Since it now has a private section, better turn it from a `struct` into a `class`.

throwException:
> Could you please point me to the place where this pointer is de-referenced? I somehow fail so see it and assumed that the native context never leaves the evades the scope of the `entry` method.

The `_native_context` is used in `Rpc_entrypoint::_alloc_rpc_cap`, `Rpc_entrypoint::_free_rpc_cap`  as well as `Rpc_entrypoint::_manage`.

nfeske:
Thanks Stefan for bringing me on the right track. :-)

throwException:
> Even though the patch is already in a very good shape, I still have a few remaining remarks:

These improvements are now also made.

nfeske:
The patch looks excellent now.

Upon cherry-picking the commit to staging, however, I noticed that some file headers - the ones newly added files - miss the `Copyright (C) <year> Genode Labs GmbH` notice. They solely feature a copyright notice for gapfruit. This is a problem for us, as I commented in https://github.com/genodelabs/genode/pull/3581#issuecomment-566076480. This applies also to some parts of the other two patches.

throwException:
> Upon cherry-picking the commit to staging, however, I noticed that some file headers - the ones newly added files - miss the `Copyright (C) <year> Genode Labs GmbH` notice. They solely feature a copyright notice for gapfruit. This is a problem for us, as I commented in [#3581 (comment)](https://github.com/genodelabs/genode/pull/3581#issuecomment-566076480). This applies also to some parts of the other two patches.

Thanks for the hint. I rechecked all files and added it where it was missing.

nfeske:
Thanks for the quick response. I added your commit as https://github.com/genodelabs/genode/commit/612201026cf07d3d670aceb61000836b62ab9f1f to staging, fixed a typo in the commit message, and added a reference to this pull request.

Letting our automated tests chew on it... :-)

Given that the other two patches introduce pretty substantial changes and the next Genode release is around the corner, I'm a bit afraid of disrupting the otherwise stable state of the master branch so shortly before the release. Would you be fine with us merging the other two commits shortly after the release? I'd still continue with the review just now.

nfeske:
Looking at the second commit (IPC caps):

* The patch looks already very nice. What I like most is that it largely looks like a simplification, with all the socket_pair-related code dropped.

* Is the huge list of includes needed in the tiny core_rpc_cap_alloc.cc file? I sense that this list can be largely pruned. The same applies for base/rpc_cap_alloc.cc.

* May you review your patch regarding the unneeded use of `Genode::`. E.g., I spotted a few instances in ipc.cc (`Genode::raw` should be `raw`).

With these minor details addressed, I'd be perfectly fine with taking the patch to staging. But since the ramifications regarding our big arsenal of run scripts are uncertain to me (the commit message mentions the ulimit topic), I would feel better to slightly defer the merge. What do you think, @throwException?

throwException:
In case someone was wondering: I forgot to (re)add a file. It's fixed now.

throwException:
>     * Is the huge list of includes needed in the tiny core_rpc_cap_alloc.cc file? I sense that this list can be largely pruned. The same applies for base/rpc_cap_alloc.cc.

Very true, I managed to reduce these includes.

>     * May you review your patch regarding the unneeded use of `Genode::`. E.g., I spotted a few instances in ipc.cc (`Genode::raw` should be `raw`).

Actually the namespace is needed there, otherwise `raw` conflicts with `Native_capability::raw()`. But I reduced `Genode::` in some other places.

> With these minor details addressed, I'd be perfectly fine with taking the patch to staging. But since the ramifications regarding our big arsenal of run scripts are uncertain to me (the commit message mentions the ulimit topic), I would feel better to slightly defer the merge. What do you think, @throwException?

The ulimit question certainly needs to be addressed. It might be a good idea to set the ulimit in `tool/run/power_on/linux` to value configurable from the runscript and, if needed, inform the user on how to upgrade it in typical distros.

Otherwise I'm pretty confident about not breaking things since the depot autopilot tests work fine. I certainly hope to see the changes in the upcoming Genode release.

nfeske:
> Otherwise I'm pretty confident about not breaking things since the depot autopilot tests work fine. I certainly hope to see the changes in the upcoming Genode release.

It is very good to know that depot autopilot is happy with your changes!

I'm mostly worried about corner cases related to lx_hybrid components and about interactive scenarios such as ports/run/bash.run or gems/run/leitzentrale.run. Those and several other run scripts need manual testing that I'm not able to pursue right now. Even if I did, should I encounter problems, I am unable to troubleshoot them within the next week either. Also, I have no tangible feeling for possible performance implications. I would feel much better giving such a cross-cutting feature the manual testing exposure it deserves before pushing it onto the Genode users at large. The time until the release is too short for that.

It would be a pity to rush the feature into the release and then notice breakage, performance regressions, or Linux-distro compatibility issues shortly after the release is out.

I wonder, why would the inclusion of the feature in the release 20.02 be in some way critical for you? Let's better strive for having it in the master branch shortly after, and deliver it in a time-tested fashion in 20.05. Less uncertainty. Less stress.

throwException:
> I wonder, why would the inclusion of the feature in the release 20.02 be in some way critical for you? Let's better strive for having it in the master branch shortly after, and deliver it in a time-tested fashion in 20.05. Less uncertainty. Less stress.

No I don't think its critical. I was just hoping to avoid too many conflicts with future work of others as this pull request is quite involved. Maybe you could integrate this pull request into master shortly after the upcoming release? This way everyone can use a well tested release but people are also on notice about the changes.

nfeske:
@throwException Thank you for your understanding.

> Maybe you could integrate this pull request into master shortly after the upcoming release?

That  is definitely my plan!

nfeske:
I just tried to cherry-pick https://github.com/genodelabs/genode/pull/3581/commits/9244db041a6d3ea75208114d8a50a16dbf8390ca to staging while applying a few cosmetic tweaks (https://github.com/nfeske/genode/commit/1db3400323df8919b0ad74ac08b2523b128c00c0). I wanted to have our autopilot have a look while diving into the new mechanics and pursuing manual testing.

While doing the latter, I observed the following problem with executing the demo.run script:

```
Genode sculpt-20.02-10-g3c64e1c002 <local changes>
17592186044415 MiB RAM and 8997 caps assigned to init
[init] Warning: scout assigned caps (10000) exceed available caps (6628)
Kernel: 24573:24576 lx_socketpair failed with -24
Error: Uncaught exception of type 'Genode::Ipc_error'
Warning: abort called - thread: entrypoint
Kernel: 24587:24601 lx_sendmsg to sd 7 failed with -111 in lx_call()
Kernel: 24589:24598 lx_sendmsg to sd 19 failed with -111 in lx_call()
Warning: blocking canceled in entrypoint constructor
Kernel: 24573:24576 lx_sendmsg failed with -9 in lx_reply() reply_socket=17743240
```

It does not look always the same. Sometimes it looks as follows:

```
Genode sculpt-20.02-10-g3c64e1c002 <local changes>
17592186044415 MiB RAM and 8997 caps assigned to init
[init] Warning: scout assigned caps (10000) exceed available caps (6628)
Kernel: 25835:25838 lx_socketpair failed with -24
Error: Uncaught exception of type 'Genode::Ipc_error'
Warning: abort called - thread: entrypoint
```

@throwException Could you please have a look?

chelmuth:
My errno to strerror tool (https://gist.github.com/chelmuth/953a75201bab1106172f9846ea83d1e7) reveals `errno 24 is "Too many open files".`

throwException:
@nfeske That looks like the file descriptor limit. What's your `ulimit -n`?

nfeske:
@throwException that was the issue indeed! Thank you for pointing it out. The value was 1024. After `ulimit -n 4096`, the demo scenario starts fine.

Now, I'm thinking about ways how to bring the user on the right track whenever this issue occurs.

I remember your suggestion about letting the run tool adjust the limit to the max automatically. Given that the limit strikes now quite early, we should definitely add that feature to the run tool.

nfeske:
I continued testing a bit more. The good news is that bash.run starts up successfully. Also fork/execve seems to work in principle (I feared that this might break). However, when repeatedly issuing `ls` (4 times), the following error occurs:

```
Kernel: 6030:6034 lx_sendmsg failed with -9 in lx_reply() reply_socket=17743240
```

Since `ls` works at first, this behavior hints at some kind of file-descriptor leak. While looking a bit closer, I noticed that core's /proc/<pid>/fd/ directory fills up over time, even when not issuing any command.

throwException:
> Since `ls` works at first, this behavior hints at some kind of file-descriptor leak. While looking a bit closer, I noticed that core's /proc//fd/ directory fills up over time, even when not issuing any command.

I can confirm that its a file descriptor e.g. capability leak. I found out that it stems from capabilities which are transferred (`insert_sds_into_message`) and suspect the timer but don't understand why the timer would periodically transfer capabilities to another process (via core). As I understand this mechanism it's only use to initiate as connection/session.

nfeske:
On base-linux, each signal is submitted by issuing a `Pd_session::submit` RPC function at core's PD session. This RPC function takes the signal-context capability as argument.

Hence, each time when the timer driver wakes up one of its clients, this RPC function is called at core, which prompts core to notify the respective client.

throwException:
> On base-linux, each signal is submitted by issuing a `Pd_session::submit` RPC function at core's PD session. This RPC function takes the signal-context capability as argument.

I found the problem: The method `Signal_delivery_proxy_component::_deliver_from_ep` does not free the capability e.g. file descriptor received from the other thread.

I'm unsure how to fix that in a senseable way. Also other RPC function might also not free the capabilities they received (and do not store for later use.)

nfeske:
Capability objects work similar to reference-counted smart pointers. Whenever the ref count reaches zero, the kernel selector can be freed. On kernels like nova, base-hw, foc, or sel4, this mechanism is routinely at play. In contrast, on base-linux, this was not needed so far because there was no 1:1 relationship between a Genode::Capability and a FD. So reaching ref count of zero did not require such a side effec. With your patch, base-linux becomes much closer to the capability-supporting kernels. So the ref count reaching zero must be handled by closing the FD.

@throwException I cannot pinpoint the missing code snippet immediately, but maybe you can follow the calls of dec_ref on base-linux and compare the behavior to, let's say sel4?

throwException:
> @throwException I cannot pinpoint the missing code snippet immediately, but maybe you can follow the calls of dec_ref on base-linux and compare the behavior to, let's say sel4?

I was finally able to create a version which does not leak file descriptors by itself in the `run/bash` scenario.

For the moment, it's an extra commit to make it easier to inspect. Also it's not yet clean code and duplicates way to much code from `base` to `base-linux` my liking.

@nfeske Which way would you like to go with code deduplication?

nfeske:
Great that you got the leak covered!

Today I got around with looking closer at your fd-lifetime patch. I liked seeing that you could replace the fds by the Native_capability types at several places.

The code duplication is unfortunate indeed. Although it cannot be completely avoided, we should at least strive to not duplicate the `Capability_space_tpl` header because it is really almost identical. The other duplications, I could (at least temporarily) live with. So let us concentrate on that one:

Besides the trivial things like removing the include of `linux_syscalls.h` or adding `Noncopyable` (good catch!) to the generic version, you apparently missed suitable kernel-specific hooks needed during the import and once the ref count reaches zero. So let us refine the generic version with the hooks needed - but doing this as a separate commit (preceding the others) because it touches non-base-linux parts.

For the import, the offending `_lookup` method could probably be moved into a platform-specific `Linux_ipc_cap_data` constructor (not using the `Ipc_cap_data` type found in the header but a Linux-specific one defined outside this header). Similarly, the hook for the condition where the ref count reaches zero could be the `~Linux_ipc_cap_data` destructor. All that's missing from the generic code is the call of the destructor inside `dec_ref` method.

I sense that introducing the `Linux_ipc_cap_data` type (in a separate header) and instantiating the `Capability_space_tpl` with our new type as second argument will almost do the trick. Do you think this would be reasonable idea to try?

nfeske:
> Similarly, the hook for the condition where the ref count reaches zero could be the `~Linux_ipc_cap_data` destructor. 

After thinking about it some more, we should better introduce a method `DATA::release` instead of using the destructor. This would make its execution of this code more explicit.

throwException:
> For the import, the offending `_lookup` method could probably be moved into a platform-specific `Linux_ipc_cap_data` constructor (not using the `Ipc_cap_data` type found in the header but a Linux-specific one defined outside this header).

I don't see how to achieve this yet, as the `_lookup` accesses the `Capability_space_tpl::_caps_data` data structure. Also, as an added twist, we discovered the last implementation was vulnerable to race conditions and needed to use the lock.

Also, the `Capability_space_tpl::lookup` and `Capability_space_tpl::import` probably also requires a lock since a possible change in data structure between `_lookup` and use would create another race condition.

In addition we discovered that `Native_capability::operator==` did no longer work as intended, because the Linux kernel would sometimes (when?) hand out new file descriptor numbers when a capability is transferred from one thread to another within one process. This necessitated a platform-specific version of the overloaded operator.

nfeske:
In the light of all these realizations, it probably the best to drop the use of the generic `Capability_space_tpl`, which was only used on platforms w/o kernel-protected capabilities so far.

nfeske:
@throwException Do you have further changes in the pipe or is the current state meant as my starting point for further testing and the merge to staging now? In the latter case, I will probably need to ponder further about ways for reducing the duplicated code, getting my hands dirty myself (instead of merely giving suggestions).

throwException:
> Do you have further changes in the pipe or is the current state meant as my starting point for further testing and the merge to staging now?

Not unless another problem with the current implementation creeps up.

> In the latter case, I will probably need to ponder further about ways for reducing the duplicated code, getting my hands dirty myself (instead of merely giving suggestions).

@nfeske That would be much appreciated, if you can spare the time. For us, the important thing is to get seccomp and capabilities in base-linux production-ready as soon as possible. So please let me know if you intend to take the drivers seat. I'd be happy to help out and we'll be conducting further testing in any case.

throwException:
> > Do you have further changes in the pipe or is the current state meant as my starting point for further testing and the merge to staging now?
> 
> Not unless another problem with the current implementation creeps up.

@nfeske And sure enough, a problem has shown itself: The performance is not all that good, especially on ARM. I'll try to improve it now.

nfeske:
I'll turn to further manual testing next week and - if everything looks bright - I'll (maybe temporarily) put it on staging to see how all the automated tests beyond depot-autopilot fare. It goes without saying that I am going to report back.

> The performance is not all that good, especially on ARM. I'll try to improve it now.

Thank you for your continuous effort!

nfeske:
@throwException I picked up your current branch for further testing and refinement.

Along this way, I solved the rlimit problem via #3721, which restores the original experience of executing run scripts like demo.run without manually tweaking the ulimit. A also rebased your work on the current staging branch, including the adaptation to the new `Mutex` API.

I can confirm that the bash.run script works now and does not leak file descriptors anymore. Once running, the number of open FDs at core remains at around 1380. To my surprise, core stays at 15-20% CPU load now (according to output of `ps aux`). Before the change, its load was below 1.5%. We should investigate where this busyness comes from.

Further testing of the tool_chain_auto.run script was successful too, which is really great! But here, the performance degradation you just mentioned becomes quite visible. On my machine, the duration of this particular test degraded from 194 seconds to 550 seconds.

Finally, I could successfully run the leitzentrale.run script, which is by far the most complex scenario we have for base-linux. That's fantastic! That said, the performance degradation becomes very much visible in this scenario. The interactive performance suffers.

I'm very happy that the current version passes those sophisticated manual tests. That is a very impressive achievement.

With regard to merging the current state, I have two concerns:

* The performance degradation is quite noticeable and may be perceived as a regression for the regular use case of base-linux during development. So for this common case, I am afraid that the change may not be for the better.

* I have to work on reducing the duplication of code and other parts that I am not completely happy with. E.g., the nested locking scheme in capability_space_linux.h does not look good to me.

I have made my current working branch available at https://github.com/nfeske/genode/commits/merge_base_linux

While I will take care of the second point, I would very much appreciate a further investigation from your side regarding the performance problem.

throwException:
>     * The performance degradation is quite noticeable and may be perceived as a regression for the regular use case of base-linux during development. So for this common case, I am afraid that the change may not be for the better.

I'm quite sure I know already where a substantial part of the performance degradation comes from: The _lookup in src/lib/base/rpc_cap_alloc_socketpair.cc function which linearly runs through all of the caps and is invoked on every RPC. I also have an faster version using a simple lookup array.

The trouble is that make the implementation faster has unearthed a race condition I cannot seem to pinpoint. Maybe you can take a look at that.

nfeske:
@throwException Thanks for the reverse-lookup optimization. However, a recursive lock can never be a solution. Let me try to simplify the synchronization scheme in a different way.

throwException:
> Let me try to simplify the synchronization scheme in a different way.

That would be greatly appreciated. I couldn't think of a simplification that keeps the general structure of the capability accounting.

nfeske:
> That would be greatly appreciated. I couldn't think of a simplification that keeps the general structure of the capability accounting.

How do you think about https://github.com/nfeske/genode/commit/30c0f0160c25752ec0b49a657dc149f6afa9b4cc?

You mentioned a race condition. Can you give me a hint how I could observe that?

nfeske:
@throwException While looking at your optimiation patch, I got the sense that your optimization replicates the purpose of the existing AVL tree (of `Tree_managed_data` nodes). I wonder, why aren't you using the existing mechanism in the first place? Shouldn't there be a one-to-one relationship between the `input_sd` and the `Rpc_obj_key`? The use of the AVL tree has the advantage over your array that it is already allocated as part of the `_caps_data` array.

throwException:
> You mentioned a race condition. Can you give me a hint how I could observe that?

I obeserved it be running the depot autopilot. Sometimes (actually quite infrequently) tests fail in one of several failure modes:

* ``lx_sendmsg` returns -111 in `lx_call` (I assume the remote entrypoint has prematurely closed the output socket)

* `lx_sendmsg` returns -9 in `lx_replay` (I assume somehow the input_socket got close prematurely)

* test gets stuck without error message

throwException:
> @throwException While looking at your optimiation patch, I got the sense that your optimization replicates the purpose of the existing AVL tree (of `Tree_managed_data` nodes). I wonder, why aren't you using the existing mechanism in the first place? Shouldn't there be a one-to-one relationship between the `input_sd` and the `Rpc_obj_key`? The use of the AVL tree has the advantage over your array that it is already allocated as part of the `_caps_data` array.

I thought so too (even tried it out), but I came to the conclusion that `Rpc_obj_key` must contain the output_socket of the process for RPC objects to be addressed correctly.

throwException:
> I thought so too (even tried it out), but I came to the conclusion that `Rpc_obj_key` must contain the output_socket valid in the hosting process for RPC objects to be addressed correctly.

(In other cases, the `Rpc_obj_key` contains the local address of an object.)

One would acutally need two AVL trees with the same elements but different keys to achieve both lookup functions at the same time.

nfeske:
Commit https://github.com/nfeske/genode/commit/4ed8d283413ec2cf211ecfcc5765efd9dd97c462 removes the local socket from the `Rpc_destination`. Tested with the demo.run and bash.run scripts. I'll now proceed trying to use the existing AVL tree for the lookup.

nfeske:
Upon close inspection of your custom lookup function, it does not seem to work as intended.

Its primary purpose is the lookup of RPC objects for capabilities given as RPC arguments (not the lookup of the invoked cap because this is handled already by using the local sd as badge). For the former purpose, the lookup function always returns `nullptr`. This is logical because when a capability of a server re-enters to the server as argument, the Linux kernel hands out a new socket descriptor number. Hence, the lookup always fails and a new capability is inserted into the capability space with the client-provided key as `Rpc_obj_key`. This is wrong on two accounts.

* The import should draw the connection to the already existing entry of the server's capability space and return the existing entry. Otherwise multiple `Native_capability` objects would end up pointing to the same RPC object but via different keys. (that said, this does not create a big problem)

* More importantly, we cannot trust the client-provided batch values as found in the message for the capability arguments as they are transferred as plain data. So the key can be forged.

The bottom line is that the mechanism works only because no client misbehaves and the lookup function always returns nullptr for the lookup of capability arguments.

I suppose you reached this implementation to accommodate the use of the import function in the path of `ipc_reply_wait` for the construction of the reply capability for the `Rpc_request`. This can be avoided altogether by storing the reply socket in the context of the entrypoint similar to how this is done on seL4.

I will therefore remove the use of `import` from the ipc_reply_wait and subsequently try to find a way to make the lookup work as intended. The latter comes down to the question of how to compare whether two socket descriptors refer to the same kernel object. In the original implementation I used to compare the names returned by `getpeername`, taking the client-provided key as a mere hint.

nfeske:
The `Native_context` does not seem to be very large anymore. As its role diminished with your change from `select` to `epoll`? If yes, we should roll back the notion of the `Native_context` and store its tiny members as part of the existing `Native_thread`. @throwException Could you clarify?

throwException:
> The `Native_context` does not seem to be very large anymore. As its role diminished with your change from `select` to `epoll`? If yes, we should roll back the notion of the `Native_context` and store its tiny members as part of the existing `Native_thread`. @throwException Could you clarify?

Yes, the `Native_context` hold no longer a list of sockets as this is now held by the Linux kernel. I'm not sure, if `Native_thread` is the right place, since previously it was stored in `Rpc_entrypoint`.

throwException:
>     * More importantly, we cannot trust the client-provided batch values as found in the message for the capability arguments as they are transferred as plain data. So the key can be forged.

In my understanding, we don't trust the badge, but use the receiving socket descriptor which equals the badge if the client is well-behaved.

nfeske:
> In my understanding, we don't trust the badge, but use the receiving socket descriptor which equals the badge if the client is well-behaved.

That is true of the invocation of a capability. But it does not apply for the delegation of a capability as message payload.

nfeske:
> I'm not sure, if Native_thread is the right place, since previously it was stored in Rpc_entrypoint.

Either way would be fine. However, since the need for `Native_context` is gone, we should revert the introduction of `Native_context`.

nfeske:
I will take care of these changes.

nfeske:
> I will therefore remove the use of import from the ipc_reply_wait and subsequently try to find a way to make the lookup work as intended.

Just for the record, I just noticed that moving the reply capability into the `Native_thread` won't work because on base-linux, RPC requests are processed out of order in the case of signal delivery. So keeping only one reply capability per entrypoint won't be possible.

nfeske:
> But it does not apply for the delegation of a capability as message payload.

Unfortunately, this has rather huge ramifications. The current approach of using socket descriptor numbers as RPC object keys creates aliasing problems within the capability space that are not solvable without trusting the client. I remember sharing this problem with you on our whiteboard session in summer. But given its quite hard to wrap ones head around it, I may have missed to bring it well across.

A solution would be the use of the global peer names behind the SDs as keys and use of getpeername to prove the integrity of client-provided values. Other approaches (e.g., the successive translation between of distinct rpc-key namespaces) would be much more elaborate (and also less general).

I will give it a shot.

nfeske:
> Sometimes (actually quite infrequently) tests fail in one of several failure modes:

These effects can be reproduced quite reliably with the leitzentrale.run script.

throwException:
> > But it does not apply for the delegation of a capability as message payload.
> 
> Unfortunately, this has rather huge ramifications. The current approach of using socket descriptor numbers as RPC object keys creates aliasing problems within the capability space that are not solvable without trusting the client. I remember sharing this problem with you on our whiteboard session in summer. But given its quite hard to wrap ones head around it, I may have missed to bring it well across.
> 
> A solution would be the use of the global peer names behind the SDs as keys and use of getpeername to prove the integrity of client-provided values. Other approaches (e.g., the successive translation between of distinct rpc-key namespaces) would be much more elaborate (and also less general).

So, the problem is verifying delegation without having the capability invoked?

nfeske:
> So, the problem is verifying delegation without having the capability invoked?

Yes.

nfeske:
> A solution would be the use of the global peer names behind the SDs as keys...

This should better be the inode number of the remote socket, which can easily be obtained via `fstat`. I remember that you actually suggested this during out conversations in summer.

throwException:
> This should better be the inode number of the remote socket, which can easily be obtained via `fstat`. I remember that you actually suggested this during out conversations in summer.

Yes. Perhaps I tried to keep it too close to the existing implementation.

nfeske:
Commit https://github.com/nfeske/genode/commit/8430d5b51b1f26a5a62b76d49e362ba511cf4099 solves most of the problems, including the integrity protection of delegated capabilities:

- Secure the integrity of delegated capabilities, by comparing using inodes
- Use existing AVL-tree for capability lookup (fix performance issues)
- Remove obsolte code (core's rpc_cap_factory)
- Sleep infinitely instead of throwing Ipc_error
- Tested with demo.run, bash.run
- The tool_chain_auto.run script triggers sporadic errors
  "lx_recvmsg failed with -9 in lx_call()"
- Performance of leitzentrale.run seems to be back to where it
  was (when it works, sometimes it breaks with above error -9).

Further work items are

- Removing the use of `Native_context`
- Tracking down the remaining sporadic error -9
- Depot-autopilot testing
- Removing debug code (plenty of raw messages)
- Reducing the overall patch size against staging

Since I'm leaving almost no stone unturned, I'd like to squash everything into one commit under my name, of course by giving proper credits to you in the commit message. Would you be fine with that?

throwException:
> Commit [nfeske@8430d5b](https://github.com/nfeske/genode/commit/8430d5b51b1f26a5a62b76d49e362ba511cf4099) solves most of the problems, including the integrity protection of delegated capabilities:

Great!

> Since I'm leaving almost no stone unturned, I'd like to squash everything into one commit under my name, of course by giving proper credits to you in the commit message. Would you be fine with that?

Fine by me.

nfeske:
Commit https://github.com/nfeske/genode/commit/e87e3f62f149c824869b6885315e3fa1faefdd74 removes the use of `Native_context`.

nfeske:
The commit series above reduce the overall patch to the parts that are essential for the feature.

The sporadic race conditions prevail, which are observable via the leitzentrale.run script.

Next up, I will turn to the epoll functions, which happen to be not synchronized but are accessed by different threads. I cannot say that they are reason for race conditions but they make me curious.

nfeske:
Commit https://github.com/nfeske/genode/commit/8b5a09cd91e3304d8138b35416fb01ede16b5059 fixes the mysterious race-condition problems.

They were the result of doubly closing socket descriptors. Whenever the Linux kernel allocated a new SD in-between the two `lx_close` calls (one of them superfluous), the second call would close the freshly allocated SD. From the kernel's perspective, the SD is unused but in the component's capability space, it is still present. When now the kernel allocates the same SD another time, the intermediate capability would wrongly refer to the third allocated SD, triggering all kinds of funny effects down the road (like sending RPCs to the wrong servers). This was a though nut to crack.

I successfully tested the new version with bash.run, demo.run, and leitzentrale.run. Unfortunately, the tool_chain_auto.run scenario gets silently stuck at some point. So there is still some other issue left. I nevertheless think that now is good time to let our automated tests chew on the current state.

nfeske:
> I nevertheless think that now is good time to let our automated tests chew on the current state.

I (temporarily) merged the squashed commit https://github.com/genodelabs/genode/commit/4c37f99fd945f62b2fccf94045f0cf7b50312c5e to staging now to see the outcome of the automated tests.

Known remaining work items:
- The completion of the tool_chain_auto scenario, which gets stuck right now.
- The removal of debug code (such as `raw` messages, busy loops, throwing towels in the form of exceptions) once we reached a stable state.

throwException:
> I successfully tested the new version with bash.run, demo.run, and leitzentrale.run. Unfortunately, the tool_chain_auto.run scenario gets silently stuck at some point. So there is still some other issue left. I nevertheless think that now is good time to let our automated tests chew on the current state.

This is really great news. I'll get our test running, too.

nfeske:
The two fixups https://github.com/genodelabs/genode/commit/3e73a9ffe1fcd510d2e2b1e463fdf80f928c7743 and https://github.com/genodelabs/genode/commit/2dcb4fca0e5945fe2ad0ad4a6125101109f2a27c cover the winding-down of entrypoints (test-signal, test-libc_fork, test-entrypoint were affected) and 32-bit Linux (incompatibility of fstat bindings between 32 and 64 bit, many tests affected).

nfeske:
To my delight, all our automated tests succeed now on 32-bit and 64-bit, including all 84 depot-autopilot tests and about 30 run scripts executed from the autopilot.list. What a relief!

Interestingly, the performance of the tool_chain_auto test improved from 195 seconds to 105 seconds, which is quite cool. I'm not 100% sure that the tool_chain_auto test is perfectly stable since it did not succeed on my manual test run two days ago. So we will need to keep an eye on it.

That said, I feel confident to wrap up the patch by removing the debug code (as another fixup). Besides that cleanup work, the patch will go into master in the current form.

@throwException I'm curious, what is the situation on your Linux/ARM variant? Should everything be fine, it may be worthwhile for you to cherry-pick the seccomp patch on top. This one might need a few minor adaptations because of the change of the Linux syscalls used now. I'm especially referring the use of `dup` outside of core and the use of `fstat/fstat64`. I must admit that I haven't had a detailed look at the seccomp patch yet. But given that it is far less invasive than the socket-descriptor-RPC work, I'm quite positive about it.

nfeske:
> Interestingly, the performance of the tool_chain_auto test improved from 195 seconds to 105 seconds,

I have to correct myself. The 105 seconds are measured on our test equipment whereas the 195 seconds were measured on my laptop. So the comparison is not meaningful at all.

I manually ran the test on my laptop again. It succeeded (yay!) and took 203 seconds, which is a plausible value.

throwException:
> To my delight, all our automated tests succeed now on 32-bit and 64-bit, including all 84 depot-autopilot tests and about 30 run scripts executed from the autopilot.list. What a relief!

This is truly great news.

> @throwException I'm curious, what is the situation on your Linux/ARM variant? Should everything be fine, it may be worthwhile for you to cherry-pick the seccomp patch on top. This one might need a few minor adaptations because of the change of the Linux syscalls used now. I'm especially referring the use of `dup` outside of core and the use of `fstat/fstat64`. I must admit that I haven't had a detailed look at the seccomp patch yet. But given that it is far less invasive than the socket-descriptor-RPC work, I'm quite positive about it.

I agree the seccomp patch should work fine with just minor modifications in the permitted syscalls.

As for ARM/Linux I don't got around to do the testing yet, but I intend to return to this shortly.

throwException:
@nfeske I'm happy to confirm, that the depot autopilot with your latest fixups sucessfully runs on our ARM/Linux.

nfeske:
@throwException Thanks for the confirmation. So we are both happy. ;-)

nfeske:
Commit https://github.com/genodelabs/genode/commit/b134867f3101305a178931ec5c4dbf664d2d83fa reverts the introduction of `Native_context`.

nfeske:
@throwException As the release 20.05 is approaching, now would be a good time to wrap up this pull request by moving the seccomp commit to staging. Are you willing to take on the rebasing of the commit to the current staging branch?

throwException:
> Are you willing to take on the rebasing of the commit to the current staging branch?

There is a dependency on the following issue #3612 because of the `base/mutex.h`. The commit from this issue does not seem to be on staging yet. Shall I include this?

nfeske:
> There is a dependency on the following issue #3612 because of the base/mutex.h.

Oh, I was unaware of that.

> Shall I include this?

Yes, that would be very good. So I can examine both.

nfeske:
@throwException I have seen that you pushed a commit but I'm left a bit puzzled. First, I'm unable to draw the connection of how this commit relates to the seccomp feature. And second, your current topic branch is based on a quite old (by now) master branch. In order to include the feature in the release, it needs to be rebased (and tested) on top of the current version of the staging branch instead.

throwException:
> @throwException I have seen that you pushed a commit but I'm left a bit puzzled. First, I'm unable to draw the connection of how this commit relates to the seccomp feature. And second, your current topic branch is based on a quite old (by now) master branch. In order to include the feature in the release, it needs to be rebased (and tested) on top of the current version of the staging branch instead.

@nfeske: Oh, it seems I completely misunderstood. I tried to rebase the base-linux capabilities patch and noticed it was already on the current staging. Rebasing the seccomp patch makes much more sense.

throwException:
@nfeske Here is the seccomp commit rebased for staging.

nfeske:
Super! A quick try to execute the demo.run script worked perfectly. :-)

I merged your commit to staging while slightly tweaking the commit message (making is less specific to gapfruit's motivation, referencing this issue). Are you fine with these adjustments in https://github.com/genodelabs/genode/commit/f4f6e61bc031cc4aef6e0182d87361052055a0bf?

throwException:
> I merged your commit to staging while slightly tweaking the commit message (making is less specific to gapfruit's motivation, referencing this issue). Are you fine with these adjustments in [f4f6e61](https://github.com/genodelabs/genode/commit/f4f6e61bc031cc4aef6e0182d87361052055a0bf)?

That's fine.

nfeske:
The seccomp.mk library is missing for x86_32, which currently lets all builds and tests fails on this architecture. I would very much appreciate a fixup.

throwException:
> The seccomp.mk library is missing for x86_32, which currently lets all builds and tests fails on this architecture. I would very much appreciate a fixup.

@nfeske I provided a fixup for x86_32.

nfeske:
Thank you. I force-pushed the new version to staging.

nfeske:
The test results are in. They look almost perfect!

There is only one single regression, namely the thread.run test, which gets stuck with the seccomp patch but runs file with this single commit reverted. The problem occurs deterministically on both 32 and 64-bit x86:

```
spawn ./core
Genode sculpt-20.02-202-ga708972aff
17592186044415 MiB RAM and 8997 caps assigned to init
[init -> test-thread] --- thread test started ---
[init -> test-thread] destroy dependent CPU sessions in wrong order
[init -> test-thread] running 'test_stack_alloc'
Warning: blocking canceled in entrypoint constructor
```

@throwException Would you mind to take a closer look at it?

throwException:
@nfeske

> There is only one single regression, namely the thread.run test, which gets stuck with the seccomp patch but runs file with this single commit reverted. The problem occurs deterministically on both 32 and 64-bit x86:

The problem is that in `Thread::_deinit_platform_thread()` (`/repos/base-linux/src/lib/base/thread_linux.cc`) the `tgkill` syscall is called with `pid=0 tid=0`, which according to my seccomp rules isn't allowed.

I would surmise this call also can't serve any purpose and that the loop can just end if `pid=0`. Do you agree?

nfeske:
> I would surmise this call also can't serve any purpose and that the loop can just end if pid=0. Do you agree?

I'm admittedly not very familiar with this specific part of the code but I agree that the attempt to kill pid 0 does not make any sense. Leaving the loop in this case (however it occurs - that part is a bit puzzling) makes sense to me. Could you provide a fixup with only this change so that I can cherry-pick it on top of the current staging branch w/o any force-push needed?

cproc:
Maybe pid=0, tid=0 occurs in [1] where threads are created but never started

[1] https://github.com/genodelabs/genode/blob/0f27d139bd215c303f01a7491f895f97dedd8eb4/repos/base/src/test/thread/main.cc#L61-L70

chelmuth:
I'm with @cproc when looking at https://github.com/genodelabs/genode/blob/43efa03d3d9179aca38124a01ada098b943962a4/repos/base-linux/src/lib/base/thread_linux.cc#L140-L165

nfeske:
Thanks @throwException. I added https://github.com/genodelabs/genode/commit/118eaf493fc88597dada28662ff6d97f01486b73 to staging. It will be squashed with the original commit once we update the master branch.

PR #3586, opened 2019-12-17
core: avoid dangling ptr in cpu thread component

alex-ab:
if exception is triggered during construction of a thread


alex-ab:
mastered

PR #3614, opened 2020-01-22
core: add platform logging service

ehmry:
If the root child requests a LOG service with the label "platform" then
return a LOG session that logs messages directly to kernel. This is to
allow a remote test controller to recognize messages produced by an
privileged local test harness.

---

This patch makes it possible for some components to write messages directly to the serial (platform) log. This allows Genode logs to be processed by using the [Test Anything Protocol](http://testanything.org/) or Sotest.

The "platform" label was the first thing that came to mind, I'm not attached to the name.

Sotest example: https://opensource.sotest.io/results/37/381
Sotest harness: https://git.sr.ht/~ehmry/genode-sotest-producer/tree/master/src/sotest-harness/component.cc


m-stein:
Hi Emery!
I might be absolutely wrong here but isn't this what Genode::raw(args); is for?
``base/include/base/log.h``

ehmry:
I never knew about `Genode::raw(...)`, but at least on linux it prefixes output with `Kernel: `.

skalk:
Hi Emery, actually it prefixes with kernel, because it is directly
printed by it, which has no knowledge about the label to use. With
other words: with Genode::raw(...) you get direct, low-latency output,
but without the knowing which component printed the message, which is
only a problem if you have to use it cross-component-wise.

On Thu, Jan 23, 2020 at 04:29:21AM -0800, Emery Hemingway wrote:
> I never knew about `Genode::raw(...)`, but at least on linux it prefixes output with `Kernel: `.
> 
> -- 
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/genodelabs/genode/pull/3614#issuecomment-577660941

-- 
Stefan Kalkowski
Genode labs

https://github.com.skalk | https://genode.org

nfeske:
In hindsight, prefixing the messages with "Kernel:" was probably not a good idea. When seeing such a message, one intuitively thinks that the kernel has something to tell. The prefix was meant to make it easy to distinguish those messages from the regular log. Since the messages are green, this distinction is already easy enough. So I'd be fine with dropping the prefix.

Regarding the raw function, note that this debug feature is not available on all kernels. On NOVA it is not available.

Emery, does the existing `raw` function meet your needs?

ehmry:
The important thing is that some component can print to serial output without any prefix at all, escape characters included. Printing without any prefix at all seems counter to the Genode-way, but I think it works well because this would only happen in the case that core or an explicitly configured component, so really it should be quite dependable in the case of testing.

`raw` would work for me if it was free of prefix, but seems odd that it is available to arbitrary components.

If it not useful upstream I can also just sit on the patch.

nfeske:
> raw would work for me if it was free of prefix, but seems odd that it is available to arbitrary components.

That's why I asked. ;-) `raw` is indeed just a handy debug utility - one that would be disabled in a production system. In contrast, your proposal is a generic solution. I like the idea.

Implementation-wise, I'd prefer to keep a single `Log_session_component` class with a new const boolean member `_prefix` initialized by the constructor (dependent on the session label). If false, the `write` method would skip the prefixing. Intuitively, I think that this patch would be simpler.

chelmuth:
Do I understand correctly that the feature you request/implement is a kind of LOG-session attribute for configuring if the session should be labeled or not? If yes, I'd prefer to name the magic label `unlabeled` and implement it just as a bool member of `Log_session_component` (in contrast to a new class).

ehmry:
Yes, this is really about labeling or not labeling. I've pushed new patch that uses `unlabeled` as magic and reuses `Log_session_component`, but with an alternative to checking a conditional.

nfeske:
Thanks @ehmry for the new patch. I merged it to staging now.

PR #3674, opened 2020-02-29
hw: Add component for making basic ARM PSCI calls.

### LINKED ISSUE #3672: Add shared component for making arm PSCI calls.
It'd be nice to have a shared component that arm platforms can use in order to make PSCI calls.

tworaz:
So far this only exposes two functions allowing the code to turn on
additional CPUs. There is much more functionality defined by PSCI spec
but so far its not needed by Genode.

Fixes #3672


skalk:
Already outdated by recent work on issue #3672

PR #3675, opened 2020-02-29
hw: Add support for Qemu virt arm platforms.

### LINKED ISSUE #3673: Add support for Qemu arm virt machines to base-hw
My long term goal is to add VirtIO driver framework and nic, input and framebuffer VirtIO drivers to Genode. The x86_64 qemu q35 machine can use VirtIO devices out of the box. The situation on arm is a bit more complex and VirtIO devices can only be exposed for custom "virt" machines. The goal here is to allow Genode using base-hw kernel to boot on such "virt" machines.

tworaz:
This patch adds support for booting base-hw kernel on qemu-arm virt
machines. The arm_virt machine has 2GB of RAM, 2 Cortex A15 cores and
uses GICv2 interrupt controller. The arm_64_virt machine also has 2GB of
RAM, but has 4 Cortex A53 cores and uses GICv3. Both machines use PSCI
to boot additional CPU cores.

Fixes #3673


tworaz:
For the record, this patch needs PSCI component from #3672 to work.

skalk:
Already outdated by recent work on issue #3673

Issue #3299, opened 2019-04-18
platform_drv for ARM universe


skalk:
The current state of ARM device drivers for Genode is a diverse conglomeration of approaches. Most of them reference resources from board-specific definitions in `repos/base/include/drivers/defs`, although they are not necessarily bound to a specific board and its resource names. A good example here is the NXP i.MX series we broadly support. Those boards share more or less the same devices. There are even some driver examples in our codebase that depend on a board and a kernel (drivers for rpi).

In more recent drivers we tried to address this problem of re-instantiating the same driver component using different boards resp. resources, e.g., the FEC nic driver. The problem here is that the configuration of the driver is not necessarily enforced when it directly routes to the IRQ and MMIO services of core. Because core implements no policy. Therefore, it would be nice if the drivers get their resources from a middle-component like the `platform_drv` for x86.

Some Genode ARM drivers already depend on a `platform_drv`, but sadly the API for the `platform_drv` is not generic not even with regard to the same architecture. Nowadays, there are `platform_drv` implementations for: arndale  imx53  odroid_x2  rpi  x86, and each of them provide their own interfaces. Things get even more complicated if you prepare packages for those device and platform drivers, because the depot packages are principally build for an architecture and not a specific board, e.g., refer to the existent platform_drv package.

My proposal for a step out of chaos, would be the following:

Move the current x86 `platform_session` API to be a generic one. It actually aggregates the possibility to iterate over devices and to get DMA capable memory. Both aspects would be useful for ARM too. The `platform_device` on the other hand will stay architecture dependent (although the current x86 variant is more relying on PCI bus properties, which can get interesting for ARM too). For ARM we need to get the I/O memory mapped registers in form of a dataspace here, as well as the interrupt resource. Moreover, we need the ability to turn the device on and off. For the time being I think it would be ok to implicitly do this when claim/release the device.
Then we have a zoo of different information a driver might be interested in, like clock rates, framebuffer dimensions (from the rpi gpu), MII bus configurations, and so on. Because this would blow up a sane API, we need a generic way of exporting that configuration to the driver. A config rom exported by the `platform_device` naturally comes in my mind.

tomga:
@skalk could you plese answer my questions below? That would help me to verify if correctly understand current situation.

> Move the current x86 platform_session API to be a generic one. It actually aggregates the possibility to iterate over devices and to get DMA capable memory. Both aspects would be useful for ARM too. 

[...]

PR #3722, opened 2020-04-02
Add support for VirtIO drivers in Genode

tworaz:
This set of patches adds initial support for VirtIO device drivers in Genode. It first defines a new set of interfaces which VirtIO device drivers can use to locate, configure and manage VirtIO devices. The service side of these new interfaces is represented by a virtio_pci driver. The client side, aka device driver is represented by a nic driver. This driver should be fully capable of running all existing Genode networking scenarios.

In order to actually make it possible to use the newly introduced nic driver in existing qemu based networking scenarios I've did a bit of code consolidation around qemu nic setup code.

Since PCI support in Genode is right now limited to x86_64, which already has tested and working nic drivers, I've decided not to make this new nic driver a default one for x86_64 qemu machines. To use this new driver one has to manually add "virtio" to etc/specs.conf. After than scenarios like ping, fetchurl, nic_router_flood should automatically use the new driver.

In order to keep this patch set as simple as possible I've decided to not include virtio MMIO bus implementation initially. It would allow the nic driver to also work on recently added qemu virt machines. The MMIO based bus driver I've implemented does require Flatten Device Tree parser which according to @nfeske might need some consideration. I'll submit a separate PR for FDT and VirtIO MMIO support once this patch set lands on staging.

I wasn't sure how to best submit this since the patches depend on each other. If required I can submit them individually.


skalk:
Thank you @tworaz for sharing your thoughts and VirtIO work with us. I will review your pull request carefully within the next days.

skalk:
@tworaz Thanks again for your contribution! I really like to read your carefully drafted and sane code. Especially that you first cleanup the usage of NIC in Qemu scenarios within the run-scripts.

Nevertheless, I'm hesitating to take your pull request one-by-one, because of the introduction of the Virtio session interface. When reading through your commits, I've seen that you spend a lot of effort into this, which cuts me to the quick. But we try to limit the uncontrolled growth of session interfaces within Genode. In fact, one of the strengths of the framework is to have only few interfaces re-used at different places. That is why, we are very cautious nowadays to introduce new session interfaces, and better try to consolidate existing ones.

To me your Virtio session interface seems like a candidate for implementing the Platform_session/Platform_device interface. As you wrote in your commit messages, it is quite similar to the existent PCI interface, which is the Platform driver's interface currently used on x86. Please correct me if I'm wrong, or do not see specific things.

To me it seems like a good idea to consolidate the Virtio bus component together with the current consolidation work of the Platform driver (see issue #3299). Maybe we can also discuss the design in a separate issue. What do you think about this?

tworaz:
@skalk I figured this might be the case given the discussions in #3299. I fully understand your reasoning and can't say I don't agree with it. Part of the reason virtio_session is a separate interface is I started working on it back in 2019 and at that time making Platform_session/Platform_device to also cover non-x86 platforms seemed like a too big task for me to handle. Certainly not something that I could attempt to do in my spare time.

Please feel free to take parts of this PR you think are useful and discard the rest for now. Once the  #3299 is done and the dust settles I'll try to re-write my code. Until that happens I'll keep using my virtio_session interface on my private work branch ;)

skalk:
@tworaz I'm happy you take it like this. In case it gets urgent, I would cherry-pick parts from your topic branch. But I think forthe moment we can close the pull-request until the dust settled, as you phrased it ;-). Do you agree?
Apart from that, I still like to invite you to continuously share your view and needs regarding the platform driver API for a Virtio platform driver during this stage of design discussion in issue #3299.

tworaz:
@skalk I'll close this PR then. Once again thanks for the time you took to look at this code.

PR #3823, opened 2020-07-15
run: Use Qemu 5.0.0 compatile gic-version param.

tworaz:
It seems Qemu 5.0.0 no longer supports gic_version parameter.
Apparently this only worked due to some compat feature that
was dropped in Qemu 5.0.0. The replacement which works on both
old and new Qemu versions is gic-version.


skalk:
I'm fine with the pull request. I just gave it a try, and it works with older qemu 4.1.0 too. @chelmuth do you mind to take it to staging?

nfeske:
Merged to staging.

PR #3824, opened 2020-07-15
platform_session: Add missing include.

tworaz:
The code uses Genode::Attached_dataspace but fails to include header
declaring this type.


skalk:
@chelmuth I'm fine with the correction. Thank you @tworaz for sending it.

skalk:
I've fixed this issue implictly while extending the platform driver. That is why I would like to close this pull request. Thanks again for the contribution!

PR #3825, opened 2020-07-15
Initial support for VirtIO device drivers in Genode

tworaz:
This is a new version of the code I initially proposed in #3722. This new version of the original patchset no longer features a dedicated virtio bus driver with its own custom session interface. It instead relies entirely on existing platform_session and platform_device interfaces. Fortunately nowadays Genode those interfaces are available on both ARM and x86_64.

Most of the code that used to reside in MMIO and PCI implementations of the VirtIO specific session interface was moved into virtio specific helpers. Those helpers are intended to be used by individual device drivers. The virtio queue helper class is transport independent. The virtio device abstraction however is closely tied to a specific transport. This patchset adds virtio device helpers for both MMIO and PCI transports. They are completely independent from each other, but offer exactly the same public API. This allows the majority of the driver logic code to be shared, regardless of the virtio transport being used. Only the initial startup code responsible for platform device discovery and initialization of Virtio::Device instance need to be different.

Qemu VirtIO device setup is not static, it depends on presence and order of command line switches. On x86_64 this is not a problem since we locate the necessary devices via PCI bus. On arm virt machines using MMIO transport the situation is not  as simple. To solve this qemu does provide a DTB binary describing machine setup via an emulated flash chip. I've used that in my original patchset, but since adding FDT support to Genode seemed to be a bit of a concern I've decided to take a slightly different approach this time. The number of potential VirtIO devices and their base addresses are hard coded in qemu sources. We just don't know if each of those devices is present or what type of device it is. We can however probe this at runtime by examining each device registers. Such probe can yield enough information to allow us to produce a sensible config for existing arm platform driver. This is done by a virtdev_rom driver introduced by one of the patches.

Similar to my previous attempt this patch series contains a fully functional NIC driver implementation based on the new APIs I've introduced. In the current form its known to work on x86_64 pc machine and arm_v7a qemu_virt. It can also be made to work on qemu_virt arm_v8a. The problem is I have no clue how to convince Genode build system to allow generic arm platform driver to be also used on arm_v8a/qemu_virt. It currently has "REQUIRES = arm" in its target.mk file which apparently implies its arm_v7a only. To test my code on arm_v8a I've basically copied all arm platform_drv code into arm_64 subdirectory of os/src/drivers/platform. This works but is obviously not a correct solution. @skalk I hoped you'll have some idea how to solve this. It'd be said if generic arm platform dirver was limited only to armv7.


skalk:
@tworaz: thank you so much for this great contribution. Your description sounds very promising. However, I'm on the go to summer holidays. That's why I cannot review the code immediately, but in about three weeks.
To shortly answer your question regarding arm_v8a and arm. Having two different architecture target.mk files is no concern. You couild reference the very same compilation units, but only double the target.mk Makefile. But the platform driver API is the more urgent problem, because it is by now architecture dependent. A short solution is to add headers to the arm_v8a include directory that include the arm variant. The midterm solution is to get rid of the differences of the x86 and arm API, consequently having only one generic API left.

tworaz:
@skalk Thanks for the hint. I hoped there would be a nicer solution to arm/arm_v8a platform_drv problem, but this should also work. At some point I'll push a separate patch doing this, I don't want to introduce any more chagnes in this series.

Please take as much time as you need to review this. I'm not in a hurry :)

skalk:
@tworaz: I've took the time to look into your commit series more carefully, and play around with it. Let me first tell you how enthusiastic I was when reading it. It is a pleasure to see how you used Genode's abstractions in an apparently natural fashion!

Having said this there are still some roadblocks, which hindered me to merge it right now into upcoming release 20.08.
The most pressing one is: we still use a too old Qemu version in our test infrastructure to support the non-legacy viortio-mmio devices. Therefore, we either would add driver packages untested, or even would break the test-runs for virt_qemu in general.
We are willing to change this quite soon after the release. But it is too hot to change it right now.
 
Another minor issue is the introduction of additional SPEC variables. It is not your fault! Genode apparently uses those variable, however we eliminated the excessive usage of them in the past for several reasons (and still do so). The rank growth made it difficult to track how a component gets assembled. More relevant: Genode's nowadays favored depot package management does not align with the usage of SPEC variable very well - apart from architecture itself. To sum it up as a rule of thumb: use architecture-only relevant SPEC variables please.
To circumvent the additional usage of virtio SPEC variables, we need to add a virt_qemu target for x86 to test the virtio-pci devices right now. Actually, a plan I would like to follow.

Regarding the introduction of the virtio_rom component. I really like how you solved the issue of dynamic platform devices by probing and creation of a config rom. Nevertheless, although it perfectly aligns with Genode's component approach, I think it is not really necessary. The way we call qemu is under our own control, and I think we can enforce the same device settings for virt_qemu under all circumstances. Thereby we do not have changing i/o resources. For instance we could always start qemu with different device models, whether they are used or not. Do you agree, or do I oversee something? Alternatively, when playing with your commits using different qemu versions, I was using this commandline switch: "-device virtio-net-device,bus=virtio-mmio-bus.0 ...". I wondered isn't the number behind virtio-mmio-bus telling the qemu the position of the mmio device?

I've prepared an issue branch for you with a stripped down version of your commits using the virtio api and driver packages, but without the virtio-pci device package. It can be used to drive AARCH32 and AARCH64 when having Qemu >= 4.2.
If you are ok with it, I would favor to integrate that in the first place, as soon as our testing infrastructure qemu got updated.
After that I would like to wait until I put pci-discovery into the current arm platform driver and unified it with x86. Then it is a good point in time to consolidate the qemu driver landscape in between x86 and arm. I hope this is in line with your own plans, and you understand my reasons.

Thanks you again for your great contribution, it is very much appreciated!

skalk:
I forgot to refer to the topic branch: https://github.com/skalk/genode/tree/virtio_nic

tworaz:
Thanks for finding the time to look at the code.

Don't worry about landing the code before 20.08 release. I generally follow any specific release. My work branches are usually based on master, or sometimes even staging where some interesting change land there (ex I switched from master to staging when Norman replaced Framebuffer with Capture sessions recently). Also I'm using pretty recent versions of Qemu so its good that your test infra also checks older releases.

Regarding x86 and PCI its generally something I have relatively low interest in. I implemented it since I felt this would be valuable setup to test the common driver code, but nothing more besides that. I'm mostly interested in arm/arm_64 without PCI support, so please feel free handle this part of the patchset as you see fit.

I did not know SPEC variables are considered a relict of the past and something that should generally be avoided. It might be a good idea to document this somewhere.

The automatic detection of virtio devices is something that is probably more convenient for me than for you guys. I have some additional drivers that I would like to submit at some point (keyboard, mouse, framebuffer) and those can obviously be easily hard coded in proper places once the drivers get submitted. Unfortunately I also have some drivers that are unlikely to land in the main repository. With virdev_rom I don't have to touch any of the config files to locate those devices on Genode side. Still even if the component doesn't get merged into main repository I can maintain it on my end. I mostly have my own run scripts for my own stuff so I can easily insert virtdev_rom component into my test scenarios.

Your branch looks fine to me please feel free to merge it whenever you feel its ready. The only question I have is what to do about the PCI stuff, will you handle it when the time is right, or should I handle it at some point?

nfeske:
> I did not know SPEC variables are considered a relict of the past and something that should generally be avoided. It might be a good idea to document this somewhere.

SPEC variables won't go away but their scope is narrower today than in the past. The move away from the SPEC mechanism was motivated by the unification of build directories and the introduction of the kernel-agnostic ABI. It is not a clear-cut change but rather a slow transition. You can find it described in more detail here: https://genode.org/documentation/release-notes/19.05#Unified_build_directories_for_ARM

skalk:
> The automatic detection of virtio devices is something that is probably more convenient for me than for you guys. I have some additional drivers that I would like to submit at some point (keyboard, mouse, framebuffer) and those can obviously be easily hard coded in proper places once the drivers get submitted. Unfortunately I also have some drivers that are unlikely to land in the main repository. With virdev_rom I don't have to touch any of the config files to locate those devices on Genode side. Still even if the component doesn't get merged into main repository I can maintain it on my end. I mostly have my own run scripts for my own stuff so I can easily insert virtdev_rom component into my test scenarios.

Well, then I think we can also live with the minor additional complexity when the flexibility is needed here. I can re-build the commit series to use the virtio_rom.

> Your branch looks fine to me please feel free to merge it whenever you feel its ready. The only question I have is what to do about the PCI stuff, will you handle it when the time is right, or should I handle it at some point?

I'll keep track of the PCI stuff. Thank you for your feedback!

skalk:
@tworaz I've re-worked the commit series after updating our buildbot infrastructure to use a more recent Qemu version,
re-added the virtdev_rom platform driver config generator for convenience reasons. I changed its sematic a bit to meet recent changes in the device_rom generation of the platform driver, for example I removed the io_mem offset workaround in it, because the platform driver now exports such information.
I extended the run-script adaptions (append_qemu_nic_args) to meet all run-scripts, and added missing support for qemu_virt on AARCH64 as well. In my eye the pull-request is now ready for staging.
@nfeske @chelmuth please have a look at my merge_to_staging branch, whether you agree with the commit series.

nfeske:
Thanks @tworaz and @skalk for the great series of commits! Thanks for paying so much attention to the details such as including documentation, adding an xsd file, and adhering our coding style. The cleanup of the run scripts regarding the qemu nic config is also very welcome.

I merged the commits to staging, accompanied by the minor fixup https://github.com/genodelabs/genode/commit/158ca1b027d6cdb22dbcf7a8de75c2d3af15f725.

@tworaz are you fine with the commits in their current form? I'm seeking your confirmation as they will appear in the git history naming you as the author.

tworaz:
@nfeske @skalk, sorry for the late response. I had a quick look at the version of the patches that got merged to staging branch and they look fine to me. Big thanks for finding the time to look at my code and for improving it.

nfeske:
@tworaz Thank you for the confirmation. :-)

PR #3950, opened 2020-11-13
Upstream/clang fixes v2

tworaz:
This is another small collection of fixes for pretty generic and trivial issues raised by clang 11.

### DIFF
--- a/repos/base-hw/src/bootstrap/spec/x86_64/platform.cc
+++ b/repos/base-hw/src/bootstrap/spec/x86_64/platform.cc
@@ -246,13 +246,6 @@ struct Lapic : Mmio
 	Lapic(addr_t const addr) : Mmio(addr) { }
 };
 
-static inline Genode::uint64_t rdtsc()
-{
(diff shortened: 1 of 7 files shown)


nfeske:
Thanks for the next batch of fixes.

To silence the warnings about unused arguments, removing the argument name instead of adding a new line like `(void)msg;` would avoid the addition of a new (and fairly uninteresting) line.

tworaz:
@nfeske thanks for the suggestion. The problem should be fixed in the updated PR.

ehmry:
FWIW Clang sometimes produces read-only ELF segments, which ld does not handle - 8c4bb7d84838e8c01673caa8ad45a4c042ccdd11.

tworaz:
I think its not clang itself, but LLD.

nfeske:
Thanks @tworaz. I merged your fixes with a few minor style tweaks.

nfeske:
The fixes have become part of Genode 20.11.

PR #3984, opened 2021-01-12
Minor fixes for problems reported by clang vol 3

tworaz:
This is another patchset containing minor fixes for various problems reported by clang.


repos/os/include/net/tcp.h
```diff
@@ -47,13 +47,11 @@ class Net::Tcp_packet
 		uint32_t _seq_nr;
 		uint32_t _ack_nr;
 		unsigned _data_offset : 4;
-		unsigned _reserved    : 3;
```
chelmuth, line 50: I doubt this works as intended as the class reflects the bit layout of a TCP packet and at this bit location are 3 reserved bits.

tworaz: Ah, you're right. Interestingly I did have this patch on my clang branch for a while and I do run various net related scenarios with it and so far haven't encountered any failures. Anyway, I'll drop the patch for now since I haven't found a nice way to silence unused variable warnings like this. [[maybe_unused]] and __attribute__((unused)) work with clang, but produce warnings under GCC. Void casting such members in constructor works, but this class for example does not have user defined constructor at all. Void casing in random methods works, but is rather ugly.

repos/os/src/server/vmm/gic.h
```diff
@@ -151,6 +151,8 @@ class Vmm::Gic : public Vmm::Mmio_device
 					Mmio_register gicr_igroupr0 { "GICR_IGROUPR0", Mmio_register::RO,
 					                              0x10080, 4, 0 };
 
+					virtual ~Redistributor() = default;
+
 					struct Gicr_isenabler0 : Irq_reg
```
nfeske, line 156: We usually inherit the `Interface` (https://github.com/genodelabs/genode/blob/master/repos/base/include/util/interface.h) in cases like this instead of providing a virtual destructor. It should have the same effect but is easier on the eyes.

@tworaz could you give this a try with clang?

nfeske:
I merged all commits except for https://github.com/genodelabs/genode/pull/3984/commits/cd2862c4902fe89a2c3b8904160ede0c643a70d5 to staging right away.

repos/os/src/server/vmm/gic.h
tworaz: @nfeske Inheriting Genode::Interface works like charm. Should I push fixed patch as a separate PR, or just force push this patchset with the problematic patch amended?

On a slightly related note yesterdays checkout of clang 12 also warns that Irq_reg::read and Irq_reg::write functions defined in the same file on lines 107 & 108 are missing override keywords. I'll push a separate PR for this.

nfeske: A force push from your side would be very convenient. ;-)

tworaz: Done. I've also amended commit description to reflect the new contents of the patch.

nfeske: Thank you. I cherry-picked your commit to staging.

PR #3985, opened 2021-01-12
Minor fixes for problems reported by clang vol 4

tworaz:
Yet another set of patches for various problems reported by clang. All the patches have been made while building run/wm and run/nano3d scenarios.

### DIFF
--- a/repos/gems/include/nano3d/sqrt.h
+++ b/repos/gems/include/nano3d/sqrt.h
@@ -12,7 +12,7 @@
  */
 
 #ifndef _INCLUDE__NANO3D__SQRT_H_
-#define _INCLUDE__NANO3D__SQRT_H__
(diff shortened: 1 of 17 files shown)


repos/gems/include/nano3d/sqrt.h
```diff
@@ -12,7 +12,7 @@
  */
 
 #ifndef _INCLUDE__NANO3D__SQRT_H_
-#define _INCLUDE__NANO3D__SQRT_H__
+#define _INCLUDE__NANO3D__SQRT_H_
```
nfeske, line 16: Oh, that's embarrassing.

tworaz: Its probably harmless, but its nice clang warns about problems like this.

PR #3986, opened 2021-01-12
base-linux: Drop parentheses from .cfi_undefined.

tworaz:
According to GNU as manual the syntax of this directive is:
  .cfi_undefined register

The manual does not mention the register should be in parentheses.
This works in GNU as even when those are present, but unfortunately
clang integrated-as does not parse this correctly. Both GNU and
clang's integrated assembler work fine when the extra parentheses
are omitted.


nfeske:
Merged to staging.

PR #4052, opened 2021-03-17
add I2c driver for imx8q based platform

Jadovenn:
I implemented an I2C driver component which provides a RPC interface for read/write on the bus. The driver is based on an `driver_base` interface for multiplatform support and an actual implementation for the imx8q_evk platform.

There is one open question: Currently, for accessing iomux's i2c and resistor configuration registers. I request a full page from the iomux's iomem. However, only access to i2c related registers is needed. Is there anyway to avoid requesting access to that whole page? I only need access to the registers.

### DIFF
--- a/repos/os/include/i2c_session/client.h
+++ b/repos/os/include/i2c_session/client.h
@@ -1,0 +1,54 @@
+/*
+ * \brief  I2C session client implementation
+ * \author Jean-Adrien Domage <jean-adrien.domage@gapfruit.com>
+ * \date   2021-02-25
+ */
+
+/*                                                                              
+ * Copyright (C) 2013-2021 Genode Labs GmbH                                     
+ * Copyright (C) 2021 gapfruit AG
+ *
+ * This file is part of the Genode OS framework, which is distributed
+ * under the terms of the GNU Affero General Public License version 3.
+ */
+
+#ifndef _INCLUDE__I2C_SESSION__CLIENT_H_
+#define _INCLUDE__I2C_SESSION__CLIENT_H_
+
+#include <base/rpc_client.h>
+#include <i2c_session/capability.h>
+
+namespace I2c { class Session_client; }
+
+class I2c::Session_client : public Genode::Rpc_client<I2c::Session>
+{
+public:
+
+	explicit Session_client(I2c::Session_capability session)
+	:
+		Genode::Rpc_client<I2c::Session>(session)
+	{}
+
+	void write_8bits(const Genode::uint8_t byte) override
+	{
+		call<Rpc_write_8bits>(byte);
+	}
+
+	Genode::uint8_t read_8bits() override
+	{
+		return call<Rpc_read_8bits>();
+	}
+
+	void write_16bits(const Genode::uint16_t word) override
+	{
+		call<Rpc_write_16bits>(word);
+	}
+
+	Genode::uint16_t read_16bits() override
+	{
+		return call<Rpc_read_16bits>();
+	}
+
+};
+
+#endif /* _INCLUDE__I2C_SESSION__CLIENT_H_ */
--- a/repos/os/include/i2c_session/i2c_session.h
+++ b/repos/os/include/i2c_session/i2c_session.h
@@ -1,0 +1,108 @@
+/*
+ * \brief  I2C session client implementation
+ * \author Jean-Adrien Domage <jean-adrien.domage@gapfruit.com>
+ * \date   2021-02-25
+ */
+
+/*                                                                              
+ * Copyright (C) 2013-2021 Genode Labs GmbH                                     
+ * Copyright (C) 2021 gapfruit AG
+ *
+ * This file is part of the Genode OS framework, which is distributed
+ * under the terms of the GNU Affero General Public License version 3.
+ */
+
+#ifndef _INCLUDE__I2C_SESSION__I2C_SESSION_H_
+#define _INCLUDE__I2C_SESSION__I2C_SESSION_H_
+
+#include <session/session.h>
+#include <base/stdint.h>
+
+namespace I2c {
+	using namespace Genode;
+	class Session;
+}
+
+class I2c::Session : public Genode::Session
+{
+public:
+
+	/**
+	 * \noapi
+	 */
+	static const char *service_name() { return "I2c"; }
+
+	enum { CAP_QUOTA = 2 };
+
+
+	/****************
+	 ** Exception **
+	 ****************/
+
+	/**
+	 * Execption thrown in case of a bus error
+	 * 
+	 * This exception is thrown by the driver incase of a timeout, device missing
+	 * acknoledgement and bus arbitration lost. The driver can be configured in the run script
+	 * to log descriptive messages incase of errors.
+	 */
+	class Bus_error : public Exception {};
+
+
+	/***********************
+	 ** Session interface **
+	 ***********************/
+
+	/**
+	 * Write 8 bits on the bus
+	 * 
+	 * \param byte The 8 bits to be sent
+	 * 
+	 * \throw I2c::Session::Bus_error An error occured while performing an operation on the bus
+	 */
+	virtual void    write_8bits(const uint8_t byte) = 0;
+
+	/**
+	 * Read 8 bits from the bus
+	 * 
+	 * \throw I2c::Session::Bus_error An error occured while performing an operation on the bus
+	 *
+	 * \return The 8 received bits
+	 */
+	virtual uint8_t read_8bits() = 0;
+
+	/**
+	 * Write 16 bits on the bus
+	 * 
+	 * \param word The 16 bits to be sent
+	 * 
+	 * \throw I2c::Session::Bus_error An error occured while performing an operation on the bus
+	 */
+	virtual void     write_16bits(const uint16_t word) = 0;
+
+	/**
+	 * Read 16 bits from the bus
+	 * 
+	 * \throw I2c::Session::Bus_error An error occured while performing an operation on the bus
+	 *
+	 * \return The 16 received bits
+	 */
+	virtual uint16_t read_16bits() = 0;
+
+	GENODE_RPC_THROW(Rpc_write_8bits, void, write_8bits,
+	                 GENODE_TYPE_LIST(Bus_error),
+	                 const uint8_t);
+	GENODE_RPC_THROW(Rpc_read_8bits, uint8_t, read_8bits,
+	                 GENODE_TYPE_LIST(Bus_error));
+
+	GENODE_RPC_THROW(Rpc_write_16bits, void, write_16bits,
+	                 GENODE_TYPE_LIST(Bus_error),
+	                 const uint16_t);
+	GENODE_RPC_THROW(Rpc_read_16bits, uint16_t, read_16bits,
+	                 GENODE_TYPE_LIST(Bus_error));
+
+	GENODE_RPC_INTERFACE(Rpc_write_8bits, Rpc_read_8bits,
+	                     Rpc_write_16bits, Rpc_read_16bits);
+};
+
+#endif /* _INCLUDE__I2C_SESSION__I2C_SESSION_H_ */
--- a/repos/os/src/drivers/i2c/component.h
+++ b/repos/os/src/drivers/i2c/component.h
@@ -1,0 +1,132 @@
+/*
+ * \brief  I2c session component
+ * \author Jean-Adrien Domage <jean-adrien.domage@gapfruit.com>
+ * \date   2021-02-26
+ */
+
+/*
+ * Copyright (C) 2013-2021 Genode Labs GmbH
+ * Copyright (C) 2021 gapfruit AG
+ *
+ * This file is part of the Genode OS framework, which is distributed
+ * under the terms of the GNU Affero General Public License version 3.
+ */
+
+#ifndef _I2C_DRIVER__COMPONENT_H_
+#define _I2C_DRIVER__COMPONENT_H_
+
+#include <root/component.h>
+#include <i2c_session/i2c_session.h>
+#include "i2c_interface.h"
+
+namespace I2c {
+	using namespace Genode;
+	class Session_component;
+	class Root;
+}
+
+class I2c::Session_component : public Rpc_object<I2c::Session>
+{
+private:
+
+	const Rpc_entrypoint &_ep;
+	I2c::Driver_base     &_driver;
+	const uint8_t         _device_address;
+
+public:
+
+	Session_component(const Rpc_entrypoint &ep,
+	                  I2c::Driver_base     &driver,
+	                  const uint8_t         device_address)
+	: _ep(ep), _driver(driver), _device_address(device_address)
+	{}
+
+	void write_8bits(const uint8_t byte) override
+	{
+		_driver.write(_device_address, &byte, sizeof(byte));
+	}
+
+	uint8_t read_8bits() override
+	{
+		uint8_t data = 0;
+		_driver.read(_device_address, reinterpret_cast<uint8_t*>(&data), sizeof(data));
+		return data;
+	}
+
+	void write_16bits(const uint16_t word) override
+	{
+		_driver.write(_device_address, reinterpret_cast<const uint8_t*>(&word), sizeof(word));
+	}
+
+
+	uint16_t read_16bits() override
+	{
+		uint16_t data = 0;
+		_driver.read(_device_address, reinterpret_cast<uint8_t*>(&data), sizeof(data));
+		return data;
+	}
+
+};
+
+class I2c::Root : public Root_component<I2c::Session_component>
+{
+	private:
+
+		const Rpc_entrypoint &_ep;
+		I2c::Driver_base     &_driver;
+		Xml_node const        _config;
+
+		/* address 0x0 is reserved, so if we return 0x0 it is an error */
+		uint8_t _get_device_address(const I2c::DeviceName &device_name)
+		{
+			uint8_t address = 0;
+			_config.for_each_sub_node([&address, &device_name] (Genode::Xml_node const &node) {
+				if (node.type() == "policy") {
+					I2c::DeviceName label;
+					label = node.attribute_value("label_prefix", label);
+					if (label == device_name) {
+						address = node.attribute_value("bus_address", address);
+					}
+				}
+			});
+			return address;
+		}
+
+	protected:
+
+		Session_component *_create_session(const char *args) override
+		{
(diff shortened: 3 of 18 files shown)


skalk:
@Jadovenn Thank you very much for taking the time to prepare such a nice and sane pull request.

Before I comment some details inline, I want to answer to your open question. From my point of view the configuration of the IOMUXC should not be part of the I2C driver at all. I'm not sure, whether you have any dynamic requirements regarding the IOMUXC pin settings. Given the code I don't see any dynamics. From my understanding, the IOMUXC in almost all cases has a static configuration for a given board. Nowadays, we simply pre-configure the IOMUXC within the bootstap code of base-hw. In the nearby future that code will be moved to the platform driver. Thereby it is working on other kernels too, and it will be re-configureable for the very few use-cases where this is necessary. To sum it up, if you recognized IOMUXC settings that weren't yet done on the i.MX 8MQ EVK, but the I2C controllers depend on, please apply them to `repos/base-hw/src/bootstrap/board/imx8q_evk/platform.cc` for now.

repos/os/include/i2c_session/client.h
```diff
@@ -0,0 +1,58 @@
...
+	{
+		call<Rpc_write_byte>(byte);
+	}
+
+	Genode::uint8_t read_byte() override
+	{
+		return call<Rpc_read_byte>();
+	}
+
+	void write_word(Genode::uint16_t word) override
```
skalk, line 46: I would not name it `write_word` and `read_word`, because this naming scheme is ancient and misleading, because it is dependent on the processor architecture. I think `read_16bit` and `read_8bit`, as well as `write_16bit` and `write_8bit` are much better. It is good to read and unified.

repos/os/include/i2c_session/i2c_session.h
```diff
@@ -0,0 +1,86 @@
...
+
+	virtual void             write_word(Genode::uint16_t) = 0;
+	virtual Genode::uint16_t read_word() = 0;
+
+	GENODE_RPC_THROW(Rpc_write_byte, void, write_byte,
+	                 GENODE_TYPE_LIST(Bus_error,
+	                                  Bus_timeout,
+	                                  Bus_arbitration_lost,
+	                                  Slave_error_no_ack),
+	                 Genode::uint8_t);
```
skalk, line 63: The multitude of possible exceptions here is not helpful in my eyes. A client will probably not differentiate in between all of these different error conditions. In fact from the view of a driver everything goes down to a `Bus_error` resp. IO error. I would expect a I2C bus controller driver to do everything possible to read or write a value. But if I really get into trouble, maybe because there is no slave device, it will answer with a common exception. Otherwise, the I2C driver shall be parameterized to finally succeed, or am I wrong?

repos/os/include/i2c_session/i2c_session.h
```diff
@@ -0,0 +1,86 @@
...
+	GENODE_RPC_THROW(Rpc_read_word, Genode::uint16_t, read_word,
+	                 GENODE_TYPE_LIST(Bus_error,
+	                                  Bus_timeout,
+	                                  Bus_arbitration_lost,
+	                                  Slave_error_no_ack));
+
+	GENODE_RPC_INTERFACE(Rpc_write_byte, Rpc_read_byte,
+	                     Rpc_write_word, Rpc_read_word);
+};
+
```
skalk, line 85: Please, use `/* */` in all non-temporary comments. This is part of Genode's coding style.

repos/os/src/drivers/i2c/component.h
```diff
@@ -0,0 +1,131 @@
...
+
+class I2c::Root : public Genode::Root_component<I2c::Session_component>
+{
+	private:
+
+		Genode::Rpc_entrypoint &_ep;
+		I2c::Driver_base       &_driver;
+		Genode::Xml_node const _config;
+
+		// address 0x0 is reserved, so if we return 0x0 it is an error
```
skalk, line 79: `/**/` comment see above

repos/os/src/drivers/i2c/spec/imx8q_evk/driver.cc
```diff
@@ -0,0 +1,207 @@
...
+	I2c::DeviceName device_name;
+	device_name = _config.attribute_value("device_name", device_name);
+
+	Genode::off_t bus_no = 0;
+	bus_no = _config.attribute_value("bus_no", bus_no);
+
+	Platform::Device_capability cap;
+	Platform::Device_client     device { _platform_connection.device_by_index(0) };
+
+	// iomem region for i2c control register
```
skalk, line 34: Coding style `/**/`

repos/os/src/drivers/i2c/spec/imx8q_evk/driver.cc
```diff
@@ -0,0 +1,207 @@
...
+	Genode::Io_mem_dataspace_capability i2c_ctl_mmio_cap  { device.io_mem_dataspace(0) };
+
+	// iomem region for iomux's i2c config and iomux's pull-up resistor config
```
skalk, line 37: Coding style `/**/` and the following...

Jadovenn:
> @Jadovenn Thank you very much for taking the time to prepare such a nice and sane pull request.
> 
> Before I comment some details inline, I want to answer to your open question. From my point of view the configuration of the IOMUXC should not be part of the I2C driver at all. I'm not sure, whether you have any dynamic requirements regarding the IOMUXC pin settings. Given the code I don't see any dynamics. From my understanding, the IOMUXC in almost all cases has a static configuration for a given board. Nowadays, we simply pre-configure the IOMUXC within the bootstap code of base-hw. In the nearby future that code will be moved to the platform driver. Thereby it is working on other kernels too, and it will be re-configureable for the very few use-cases where this is necessary. To sum it up, if you recognized IOMUXC settings that weren't yet done on the i.MX 8MQ EVK, but the I2C controllers depend on, please apply them to `repos/base-hw/src/bootstrap/board/imx8q_evk/platform.cc` for now.

Thank you for your explanation! It will be done

repos/os/include/i2c_session/i2c_session.h
Jadovenn: You are right. If the client receive any of this exception it can not really do anything about it. Having only the Bus_error make sens, I suggest logging some warnings where those actual exceptions are, and add a verbose parameter to the driver config to let the user decide if he wants to see those warnings or not. Does that sound good to you?

skalk: Yes that sounds perfect!

nfeske:
Thank you @Jadovenn for the very nice pull request and @skalk for the review.

I hope that you won't mind a few suggestions for coding-style-related tweaks, which will help to make the code even more consistent with Genode's code base.

* For components that merely rely on the Genode API, we usually import the Genode namespace into the component's namespace, like so:
  ```
  namespace I2c {
    using namespace Genode;
    class Session;
  }
  ```
  This eliminates the need for the countless `Genode::` prefixes and thereby improves the clarity of the code.

* Our coding style (https://genode.org/documentation/developer-resources/coding_style) facilitates the indentation of the private/protected/public class labels.

* Please pay attention to vertical whitespace. I.e., we leave two blank lines between methods/functions implemented in cc files.

* When implementing methods in cc files, we don't wrap the code in `namespace { ... }` but instead place a `using namespace` clause at the beginning of the file. This makes the indentation more consistent while also helping us to identify dead code in the code base. In particular, when declaring a function in a header file like this:
  ```
  namespace Example {

    void super_important_function();
  }
  ```
  and placing the implementation (in the cc file) within a namespace scope like this:
  ```
  namespace Example {

    void superimportant_function()
    {
      ... super important code...
    }
  }
  ```
  the compiler won't complain about the missing underscore. So it becomes easy to end up with dead (unused) code in the code base (if there is also another implementation of `super_important_function` in some other compilation unit.
  Instead, when writing the code as follows:
  ```
  void Example::superimportant_function()
  {
    ...
  }
  ```
  the compiler will produce an error about the missing declaration in the scope of `Example`.

* We try to use `const` wherever possible. The current version of the patch contains quite a few mutable variables where constants would suffice.

Jadovenn:
> Thank you @Jadovenn for the very nice pull request and @skalk for the review.
> 
> I hope that you won't mind a few suggestions for coding-style-related tweaks, which will help to make the code even more consistent with Genode's code base.
> 
> * For components that merely rely on the Genode API, we usually import the Genode namespace into the component's namespace, like so:
>   ```
>   namespace I2c {
>     using namespace Genode;
>     class Session;
>   }
>   ```
>   
>   
>   This eliminates the need for the countless `Genode::` prefixes and thereby improves the clarity of the code.
> * Our coding style (https://genode.org/documentation/developer-resources/coding_style) facilitates the indentation of the private/protected/public class labels.
> * Please pay attention to vertical whitespace. I.e., we leave two blank lines between methods/functions implemented in cc files.
> * When implementing methods in cc files, we don't wrap the code in `namespace { ... }` but instead place a `using namespace` clause at the beginning of the file. This makes the indentation more consistent while also helping us to identify dead code in the code base. In particular, when declaring a function in a header file like this:
>   ```
>   namespace Example {
>   
>     void super_important_function();
>   }
>   ```
>   
>   
>   and placing the implementation (in the cc file) within a namespace scope like this:
>   ```
>   namespace Example {
>   
>     void superimportant_function()
>     {
>       ... super important code...
>     }
>   }
>   ```
>   
>   
>   the compiler won't complain about the missing underscore. So it becomes easy to end up with dead (unused) code in the code base (if there is also another implementation of `super_important_function` in some other compilation unit.
>   Instead, when writing the code as follows:
>   ```
>   void Example::superimportant_function()
>   {
>     ...
>   }
>   ```
>   
>   
>   the compiler will produce an error about the missing declaration in the scope of `Example`.
> * We try to use `const` wherever possible. The current version of the patch contains quite a few mutable variables where constants would suffice.

@nfeske absolutely, thank you for such a detailed explanation about the coding style. I have updated the patch following your recommendation. I hope that I didn't miss anything.

skalk:
@Jadovenn thank you for incorporating the review notes! I've re-read your commit, and I'm fine with the changes. Only one thing is left, which I have overseen when initially reviewing your work: the path `repos/os/src/drivers/i2c/spec/imx8q_evk` should be changed to `repos/os/src/drivers/i2c/imx8q_evk`. The `spec` sub-directory should be used to distinguish SPEC-variable derived aspects only. The board name is not any longer a SPEC variable or aspect. Would you mind to change that path? Sorry, that I did not mentioned it the first time.

Jadovenn:
@skalk no problem, I changed it to `repos/os/src/drivers/i2c/imx8q_evk`

skalk:
@Jadovenn thank you for taking the time to incorporate our comments, and for the contribution at all! I'm looking forward to use it.
@nfeske there are no objections from my side to accept the pull-request

nfeske:
Thanks for the prompt response to my remarks. The update is very much appreciated. I merged the commit to staging, but accompanied with a fixup commit f3a4605dea1681ac265dd157198294f5d843a476 with cosmetic tweaks describes as follows.

@Jadovenn do you approve those changes?

* In our coding style, const qualifiers always appear at the right side of the type if refers to. Placing a `const` in front of a type just muddies breaks the consistency with the general rule. For example,
  ```
  const uint8_t *buffer_in
  ```
  should better be
  ```
  uint8_t const *buffer_in
  ```

* A const qualifier of a value argument in a method declaration has no effect. So it is better to omit it there. However, a const qualifier of a value argument of a method implementation makes this argument const in the scope of the method. E.g., the method declaration
  ```
  void write(uint8_t, uint8_t const *, size_t)
  ```
  is consistent with the following implementation.
  ```
  void write(uint8_t address, uint8_t const *buffer_in, size_t const buffer_size)
  ```
  The `const` of the `buffer_size` argument is technically not part of the method's signature.

* Argument names of 'override' declarations should better be skipped since they are already written out in the declaration. There is no need to repeat their names. In contrary, since the duplicated information is not checked by the compiler, inconsistencies may slip in down the road, should the name of an argument of the interface change. Hence, I changed, e.g.,
  ```
  void write(uint8_t address, uint8_t const *buffer_in, size_t buffer_size) override;
  ```
  to
  ```
  void write(uint8_t, uint8_t const *, size_t) override;
  ```
  As a plus, this makes the code more tidy.

* We use to skip the prefix `is_` from boolean values and getters

  In practice, this makes the a little more concise, e.g,

  ```
  if (_verbose) ...
  ```
  is just as descriptive as
  ```
  if (_is_verbose) ...
  ```
  but is shorter.

* Removed a couple more of superfluous `Genode::` occurrences

* Removed trailing whitespace

* I simplified the construction of `Main` in i2c/main.cc, eliminating the need to use `Constructible` for the `_driver` and `_root` members. This makes the relationship between the member variables more obvious.

* Consistent spelling of "I2C" in comments and documentation (not i2c or I2c)

* Variables that get a value assigned only once can be turned into constants. For example,

  ```
  I2c::DeviceName label;
  label = node.attribute_value("label_prefix", label);
  ```
  can better be written as
  ```
  I2c::DeviceName const label = node.attribute_value("label_prefix", label);
  ```

* The following check is not needed because it is already performed by component/root.h before calling '_create_session':

  ```
  if (ram_quota < sizeof(Session_component)) {
      warning("insufficient donated ram_quota "
              "(", ram_quota, " bytes), "
              "require ", sizeof(Session_component), " bytes");
      throw Insufficient_ram_quota();
  }
  ```

* Cleaned up list of includes in i2c_interface.h

* Don't use Cstring outside of output (e.g., log) arguments

* Renamed `DeviceName` to `Device_name`

* Removed unneeded vpath in 2c_mcp9808/target.mk

While reviewing the commit as a whole, the following remaining questions crossed my mind. Maybe you can help clarifying them?

* What is the purpose of the `_bus_mutex` in the driver? As far as I can see, the driver is not multi-threaded. So this mutex does not serve a purpose, or does it?

* Why is `_sem_cnt` marked as volatile? As far as I can see, it is used as a regular variable. Is this really a semaphore or more of a nesting level? Given the absence of multiple threads, the variable name makes one wonder a bit.

* Regarding the `_get_device_address` function, wouldn't it be appropriate to use the `Session_policy` utility defined in os/include/os/session_policy.h? In contrast to the custom `_get_device_address` code, this utility would also handle the other common ways of policy selection `label` and `label_suffix`, fostering the consistency with many other servers.

* The following code
  ```
  _config.for_each_sub_node([&address, &device_name] (Genode::Xml_node const &node) {
      if (node.type() == "policy") {
           ...
  ```
  could probably be simplified to
  ```
  _config.for_each_sub_node("policy", [&address, &device_name] (Genode::Xml_node const &node) {
      ...
  ```

  I nevertheless left it unchanged because I have no good way to test it. But maybe, you could give this change a try?

* Wouldn't it be a good idea to call the `_irq_handle` method once after registering the signal handler?

  Otherwise, an interrupt that was triggered just in-between the initialization but before the call of 'sigh' would remain unhandled. Granted, this corner case is unlikely to occur but it exists and can be prevented by calling the handle method once after the `sigh`.

If you like to refine those points, we can still add fixup commits to staging before the commit enters the master branch.

As a final question, what would be needed by me to execute the test on my i.MX8 EVK board? Do I need to connect a peripheral device to the board? If so, how must it be connected?

nfeske:
Commit https://github.com/genodelabs/genode/commit/e125d389a8777edd7b449444b63722aba6030bef contains further simplifications that I'd suggest.

- It avoids storing a copy of the `config` `Xml_node` in the `Driver_base` class because this copy is risky. The `Xml_node` contains a pointer to the underlying character data. Should we later add the ability of dynamic config changes to the driver, the stored `Xml_node` member may refer to outdated data. It is safer to capture the arguments using a dedicated self-contained type. It also keeps the driver code free from XML-parsing concerns.
- I removed the state from the `Driver_base` so the class becomes a plain abstract interface. The state is now kept only at the actual driver class. This separates the interface and its implementation more clear-cut.
- I replaced the init function and the use of `Constructible` members by plain constructor calls. This removes the constructed/not-constructed state from the picture and also avoids the uneasy looking dereferencing `->` of the constructible members (when seeing such a `->`, one always must consider that the pointer is in the wrong state).

I hope you don't feel bothered by me massaging your code. Please tell me if you support the changes, or if I'm overstepping.

Should you like the changes, I'd appreciate you giving the latest version another try on the board to make sure that I haven't inserted a regression.

Jadovenn:
Thank you, @nfeske, this is fine with me. 

Concerning your questions I did not find any reason(s) to keep the `volatile` and the `mutex` anymore, I'll remove them. Concerning the other suggestions I will address them in a new commit too and test it with your recent changes.

To run the scenario you need a sensor hardware based on the `MCP9808`. You can connect it to an imx8q_evk board using the i2c bus 2. On the schematic doc of the `imx8q_evk` you will find the connector `J801` page 11 which describe where i2c2_sda and i2c2_scl pins are. You will find GND and 3.3V pins on the same connector. You don't need to set up pull-up resistors. This is it for the hardware setup. Make sure that you update the slave address in the run-script depending on which sensor hardware you use.

Jadovenn:
@nfeske sorry but now I am confused, what are you expecting from me? Can I still commit here? (it will conflict with yours)

I did test your change, and it did not introduce any regressions.

nfeske:
> @nfeske sorry but now I am confused, what are you expecting from me? Can I still commit here? (it will conflict with yours)

My commits marked as "fixup" are meant to be squashed together to your commit once I merge the staging branch to the master branch. The sum of the work will be a single commit with you as the author.

In the meantime (before updating the master branch), each of us can come up with further fixup commits, which we can review independently.

In other words, please supply further modifications as individual patches. We  usually name such fixup commits as `fixup: "name of the original commit message"`. By mentioning the number of this pull request in your commit messages via `issue #4052`, your additional commits will automatically show up here.

Would that work for you?

nfeske:
I wonder, may you rebase your fixup commit to the current genodelabs/staging branch that already contains my fixups?

Jadovenn:
thanks @nfeske, I removed the `mutex` and `volatile`. I also replace `_get_device_address` by a `Session_policy`. I had a call to `_irq_handle` just after `sigh`.

nfeske:
Very good, thank you!

I merged your latest fixup to staging. I also added another (very minor) tweak with c1ac65dcc2156b0d14f47d8d591a045b7c2987ed, using `Genode::Interface` instead of manually declaring empty virtual destructors.

nfeske:
The feature entered the master branch just now. Thank you @Jadovenn for the valuable contribution and the very pleasant correspondence.

Jadovenn:
@nfeske @skalk Thank you very much for your reviews, this sealed my first ever contribution to an open-source project!

PR #4187, opened 2021-06-04
spi_driver: add spi driver for imx8q based board

Jadovenn:
This commit adds an interface for SPI driver. It also includes an implementation for imx8q based board that support only master mode.

### Brief

The driver expose a Spi service consumed by clients applications. Each client get access to a slave device through a session  policy defined in the driver configuration. Also, specific SPI bus setting can be configured per slave through the session's policies and updated at runtime using settings getter/setter.

A client can transfer data to a slave using the synchronous and non thread-safe `size_t transfer(char *buffer, size_t buffer_size);` RPC. The buffer content is copied to a pre-allocated buffer by the client that is then shared with the driver. The `Spi::Connection` class accept a buffer size as one of the constructor optional parameters. It allows the client application to change the shared buffer size if the default value does not suite the client needs.

### First Questions

1. Transaction 
I am wondering if it would be interesting to change the transfer signature now and use the following struct instead:
```c++
struct Transaction {
    size_t buffer_size;
};
```
The `Transaction` struct is then a "header" of the buffer that contains the buffer size and any other parameters that might be need by the driver (extended with further features, E.g  wait_state).
Do you think that might be done in this commit or can wait any further contribution if needed?

2. Spi::initialize()
Currently, the specific platform implementation is initialized and statically stored in this function. I see two other cases where others implementation would be needed for the same platform, ECSPI with DMA and Quad SPI. The driver can then decide to initialize one or another depending on the configuration that has been provided. What do you think about this? Would you change it?


repos/os/include/spi_session/client.h
```diff
@@ -0,0 +1,69 @@
...
+
+
+	/* transfer() is NOT thread-safe */
+	size_t transfer(char *buffer, size_t buffer_size) override {
+		if (buffer_size > _io_buffer.size()) {
+			throw Spi::Io_buffer_too_small();
+		}
+		memcpy(_io_buffer.local_addr<char>(), buffer, buffer_size);
+		size_t const bytes_count = call<Rpc_transfer>(buffer_size);
+		memcpy(buffer, _io_buffer.local_addr<char>(), buffer_size);
```
chelmuth, line 54: Maybe the count of bytes could be `min(bytes_count, buffer_size)` to prevent redundant copying?

Jadovenn: Yes it can, with a step back, SPI is full-duplex, so it is expected for all transaction that `bytes_count == buffer_size` otherwise an exception is thrown by the driver.

repos/os/src/drivers/spi/imx8q/ecspi/ecspi_driver.cc
```diff
@@ -0,0 +1,280 @@
...
+
+void Spi::Ecspi_driver::_bus_disable()
+{
+	_mmio.write<Mmio::Control>(0);
+}
+
+
+void Spi::Ecspi_driver::_bus_execute_transaction()
+{
+	_sem_exchange.down();
```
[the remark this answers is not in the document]
Jadovenn: I think it is not necessary, but on an initial review, it had been suggested that I added it to make sure no data race occur.  
The `_sem_exchange` is flipped by the IRQ handler while the driver's ep thread is block by `_env.ep().wait_and_dispatch_one_io_signal();`. So if it is right to you, I can replace it by an integer?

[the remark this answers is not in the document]
chelmuth: While reviewing, I discovered the use of this semaphore, which made me wondering about the requirement for multi-threading. Then, I did not find multiple threads being used and had the urge to ask ;-)

So in my opinion, it would be better to not use synchronization primitives in cases where only state/logic should be expressed. This makes the implementation more clear and also spares unnecessary overhead.

repos/os/include/spi_session/client.h
chelmuth: Please bear with me, but I'm confused by your remark. From the implementation, I got the impression that the transaction interface requires some in-buffer metadata like *this a send transaction of 100 bytes* or *please receive 200 bytes from the device*. Could you please help me understand your design (and also the relevance of *full-duplex* in this context?

repos/os/src/drivers/spi/imx8q/ecspi/ecspi_driver.h
```diff
@@ -0,0 +1,122 @@
...
+		uint8_t  clock_divider = 0;
+		uint64_t timeout       = 1000; /* in millisecond */
+	};
+
+
+private:
+
+	Genode::Env      &_env;
+	const Config      _config;
+	Timer::Connection _timer { _env };
```
chelmuth, line 52: This timer is only used as a kind of wall clock currently. Please check with [[1]](https://genodians.org/m-stein/2021-05-07-short-guide-to-the-timeout-framework) and [[2]](https://genodians.org/m-stein/2021-05-31-a-short-guide-to-the-timer-session-interface) if the use of the timeout framework (and it's `curr_time()` function) fits your requirements.

repos/os/include/spi_session/client.h
Jadovenn: So let me rephrase, the current use case is that I want to send X bytes to the slave device (and receive X bytes from it) through SPI.
 - I will use the `transfer` RPC to send those bytes by providing a buffer to it. The `transfer` client method copy bytes from the provided buffer to a swap buffer (aka _io_buffer) which is shared with the driver. (l. 52)
 - Then the proper call to the RPC is performed (l. 53).
   - The driver read from the swap buffer to a FIFO_TX stack from where words(ARM so 4 bytes) are shifted to the bus.
   - Each time, the driver write up to N bytes to the FIFO_TX, and shift them out, as much are received and shifted in into the FIFO_RX stack.
   - Those are called burst, there are from N, up to 256 bytes per burst (1 < N < 256) for the current controller.
   - After writing, the FIFO_RX is ready, and the driver immediately copies those received bytes to the swap buffer (replacing the ones who were just sent).
 -  Returning from the RPC the received bytes are now stored to the swap buffer. Then they are copied from the swap buffer to the provided buffer (l. 54). Finally, the swap buffer is ready for a new transaction. And the provided buffer contain data from the slave ready to be processed by the client.

Considering my limited experience with SPI, I might be terribly wrong about the way to interface a client with the SPI driver. Now I understand the fact that the RPC call returning `size_t const bytes_count` is confusing. Because we are expecting to read as many bytes we have written. Currently, if `bytes_count` is different from `buffer_size` the driver throw an `Spi::Bus_error` exception. So it might be a better design to have the RPC returning nothing, and the client expect an exception if not as many bytes has been read compared to how much it has sent.

skalk: @Jadovenn Thank you for your explanations! From what you've described so far, to me it seems that you map the full-duplex line protocol to the software API. Thereby, one can only do bi-directional transfers even if you only need to read or write bytes or bits. Is this correct?
The question arises whether one needs this kind of protocol on the software-level at all. From what I've read so  far, SPI is used to connect flash memory or displays, where the data-flow is one-directional at a given time. When interfacing a flash memory with your current interface proposal, one would need to copy a lot of bytes unnecessarily to e.g., read some KiB of flash memory. But there are other use-cases for SPI as well, e.g. in the FPGA communication. However, I do not see the use-case right now, why you would the full-duplex property on the software API level essentially. But this is maybe just a lack of knowledge on my side. Could you describe your use-case in more detail?

Jadovenn: The current use case is while interfacing with a TPM. There is a TPM library that prepare a TPM_cmd buffer, it forwards it to the SPI driver that give back the result to the TPM library. Then the TPM library perform further treatment on this buffer after reception.

skalk: Probably this means you send a command to the TPM via SPI, and want to receive the result immediately? Probably on the SPI link level, you would send the command, and then transfer zeros in the length of the awaited response, right? Or do you concurrently receive the result from the driver by magic? Anyway, I understand that you would held state on the driver side when only sending the command, which is probably not wanted.

To me two options seem most useful, either you collocate the SPI driver with the TPM library for the time being. Thereby, no SPI interface is needed right now, and we can collect more real-world usages before designing an interface. To me this approach seems most useful at the moment.

The alternative solution in my eyes would be to change the interface, so that it supports your use case of sending/receiving concurrently with one RPC, but also only one direction. The later approach would need a more structured way of `transfer`, which states the count of send bits (AFAIK correctly you can transfer not byte-aligned bit stream via SPI?), and how much you are willing to receive. Both lengths can be different. @Jadovenn @chelmuth Do you agree?

Jadovenn: @skalk  I do agree with both of your proposed options.  

I currently don't have a clear understanding on how to communicate with the TPM itself. Both 1 transfer to read/write or 1 transfer to read and 1 transfer to write don't work. It turns out, that initializing and communicating with a TPM through SPI is a lot more complicated than expected. My current understanding is that both cases are possible depending on the TPM's state and/or the command.  

Using a TPM as a reference device to design an SPI RPC interface for all future requirements was probably a bad idea... IMO the interface should include two "synchronous" read/write RPCs (that use 2 transfers), plus transferring and receiving concurrently in one RPC.  

I agree that waiting to have more real-world use cases and device integration needs, is the way to go for a more generic SPI interface.

skalk: @Jadovenn according to your last post, do you agree that we can close the pull-request for the time being?

Jadovenn: @skalk yes, thank you very much for the review and points that it has raised!

Issue #2309, opened 2017-03-07
Cannot get NOVA demo to run on qemu


fidget324:
I followed the directions in the Genode book, Section 2.4, to build and run the nova demo on x86_64. I'm using Genode 17.02, and qemu 2.7.1 on Ubuntu 16.04 LTS. When I run
```
$ make run/demo
```
everything seems to build just fine, and then qemu launches. The GUI for the demo is then displayed, but freezes immediately. The scout window is displayed with the green x-ray shading, but I cannot move or manipulate it at all. The entire GUI just remains frozen, and I am unable to do anything. Here's the output when I run `make run/demo`:
```
including /home/scott/Documents/genode/tool/run/power_on/qemu
including /home/scott/Documents/genode/tool/run/log/qemu
including /home/scott/Documents/genode/tool/run/image/iso
including /home/scott/Documents/genode/tool/run/boot_dir/nova
including /home/scott/Documents/genode/repos/os/run/demo.run
building targets:  core init drivers/timer server/nitpicker app/pointer app/status_bar app/xray_trigger server/liquid_framebuffer app/launchpad app/scout test/nitpicker server/nitlog drivers/framebuffer drivers/input server/report_rom server/rom_filter drivers/platform drivers/acpi server/report_rom drivers/platform/spec/x86/device_pd
spawn make core init drivers/timer server/nitpicker app/pointer app/status_bar app/xray_trigger server/liquid_framebuffer app/launchpad app/scout test/nitpicker server/nitlog drivers/framebuffer drivers/input server/report_rom server/rom_filter drivers/platform drivers/acpi server/report_rom drivers/platform/spec/x86/device_pd
make[1]: Entering directory '/home/scott/Documents/genode/build/x86_64'
Makefile:134: target 'server/report_rom' given more than once in the same rule
checking library dependencies...
Skip target drivers/framebuffer/spec/exynos because it requires exynos
Skip target drivers/framebuffer/spec/imx53 because it requires imx53
Skip target drivers/framebuffer/spec/omap4 because it requires omap4
Skip target drivers/framebuffer/spec/pl11x/pbxa9 because it requires pl11x pbxa9
Skip target drivers/framebuffer/spec/rpi because it requires rpi
Skip target drivers/framebuffer/spec/sdl because it requires linux sdl
Skip target drivers/input/spec/imx53 because it requires imx53
Skip target drivers/input/spec/ps2/pl050 because it requires pl050
[...]

PR #4282, opened 2021-10-05
Add support for keyboard and mouse input for virt_qemu boards

tworaz:
I've been using this driver for a while and decided it'd be good to finally upstream it. This should hopefully make virt_qemu boards much more usable when it comes to running interactive test scenarios like run/vim, or run/bash for example.


tworaz:
@nfeske No problem. I'm fine with the changes you've made.

WRT demo.run strange input problems, I can't see anything unexpected on my end. I'm still on Qemu 6.0, however. I'll try update to a newer version and see if it changes the situation.

nfeske:
> I'm fine with the changes you've made.

Thanks you for the feedback.

tworaz:
@nfeske I did manage to update my setup to use Qemu 6.1.0 during the weekend. Unfortunately I don't see any strange input events. I did try run/demo. run/wm, run/vim, run/nano3d. All of them seem to work as expected WRT mouse and keyboard events. I used staging branch commit bea3e8b2c2bb77566ae37df69cfab7da55c6f850 when testing this.

nfeske:
> Unfortunately I don't see any strange input events.

I suspect that the effect is similar to the one reported in https://github.com/genodelabs/genode/issues/2309#issuecomment-284860860. I'm running Qemu inside VirtualBox on top of Sculpt OS. So this may well be a problem of the stacking of two layers of virtualization. In my opinion, this is not a pressing problem since Qemu's SDL backend can easily be disabled in situations like mine. Then everything works.

chelmuth:
Added ab476d96e47 virt_qemu: use "-device ramfb"

nfeske:
This work is part of Genode 21.11. Thank you again @tworaz!

PR #4298, opened 2021-10-15
Two minor fixes for sculpt 21.10

tworaz:
First fix allows the code in reconstructible to be built with both gcc and clang.

Second one should result in potential runtime crash to be replaced with explicit error runtime message.

### DIFF
--- a/repos/base-linux/lib/mk/lx_hybrid.mk
+++ b/repos/base-linux/lib/mk/lx_hybrid.mk
@@ -2,6 +2,10 @@ SRC_CC += lx_hybrid.cc new_delete.cc capability_space.cc
 SRC_CC += signal_transmitter.cc signal.cc
 SRC_C  += libgcc.c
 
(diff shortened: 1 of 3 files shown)


tworaz:
It seems lx_hybrid uses new_delete.cc from base cxx module without allowing compiler to access libsupc++ header files. I think the easiest way to fix it is to just add STDINC=yes in lx_hybrid.mk.

There is one potential issue with this however. This is not limited to this specific case of STDINC usage. Genode toolchain in its default config builds code in hosted mode. This means that any of the standard headers shipped with GCC can access host system headers via various include_next statements. Personally I don't like that and I have some local patches that also set -ffreestanding whenever STDINC=yes is set. For maximum safety I think it'd be also good to add CC_OPT+=-ffreestanding to lx_hybrid.mk. Since Genode does not do anything like this anywhere I also did not do it for the fixup I pushed.

PR #4299, opened 2021-10-15
libm: Fix undefined __isinff and __isinfl symbol references.

tworaz:
Those symbols are not satisfied by any code or dependency of libm. As
result calling cprojf function will always crash on Genode. This crash
can be turned into link time error by adding --no-undefined to LD_OPT.

This patch provides the missing symbols by including isninf.c in libm
build.


chelmuth:
Merged to staging.

PR #4327, opened 2021-11-16
Implement native Genode VirtIO framebuffer driver

tworaz:
The goal is to use for virt_qemu boards. I wrote the driver over a year ago, but my original plan was to also extend it to cover 3d parts of the VirtIO GPU specification, so I never submitted it in a PR. I managed to get some parts of 3d accel working, but real life got in the way and I had to put the project on hold. Nowadays I once more have some spare hours I can spend on Genode, but other projects I have in mind are simply more appealing. This is why I decided to submit the driver in its current form.

Even though it lacks the virgl (3d accel) part the driver is pretty complete. It does provide the same functionality as existing ramfb driver. On top of that however virtio_fb can handle screen resolution changes. This means Qemu window can be resized, or made fullscreen and the driver will handle this as one would expect, by doing a resolution switch on VM side. In contrast ramfb only supports 1024x768 resolution and resizing Qemu window just scales the content.

Another tiny advantage of virtio_fb is that it works for both arm_v7a and arm_v8a out of the box. This is not strictly a limitation of ramfb code, it just doesn't have a spec version of target.mk for arm_v7a.


chelmuth:
I merged your commits to staging. The driver seems to work flawless with my Qemu 6.1 build but I could not interact with run/demo in a Sculpt Linux VM.

tworaz:
@chelmuth Seems like a problem with virtio input drivers submitted in #4282. Do you happen to run qemu with -display sdl? Norman mentioned having problems with it in that ticket. The current suspicion is this has something to do with SDL backend running under VirtualBox. AFAIU GTK display backend works fine in this setup. I'm using Qemu 6.1.0 on top of baremetal linux and both SDL and GTK UIs work fine for me.

I do have plans to investigate If I can run VirtualBox on Genode running on top of qemu/kvm x86_64 in order to reproduce the problem. Unfortunately I don't have and x86_64 HW I can use to run Genode natively.

chelmuth:
Thanks for the explanation. I can confirm that it works with `-display gtk`.

nfeske:
The driver is part of Genode 21.11.

Issue #4318, opened 2021-11-08
DNS settings for libc application 


sergey-platonov:
Hello all

We have noticed that for the `libc` application DNS settings will be only applied if we create the `socket` directory in the root of the `vfs` node:
```
<libc socket="/socket" ... />
<vfs>
    <dir name="socket"> <lwip dhcp="yes"> </dir>
</vfs>
```  

We used to place `socket` dir in the `/dev` and in that case, DNS does not work.

Here is a runscript that demonstrates the problem (it's merely a copy of `genode-world/run/fetchurl.inc` with few simplifications):
<details>
<summary>runscript</summary>

```
create_boot_directory

set depot_archives { }

lappend depot_archives [depot_user]/src/[base_src]
lappend depot_archives [depot_user]/pkg/[drivers_nic_pkg]
lappend depot_archives [depot_user]/src/curl
lappend depot_archives [depot_user]/src/fetchurl
lappend depot_archives [depot_user]/src/init
lappend depot_archives [depot_user]/src/libc
lappend depot_archives [depot_user]/src/libssh
lappend depot_archives [depot_user]/src/nic_router
lappend depot_archives [depot_user]/src/openssl
lappend depot_archives [depot_user]/src/vfs
lappend depot_archives [depot_user]/src/vfs_lwip
lappend depot_archives [depot_user]/src/zlib

install_config {
<config>
	<parent-provides>
		<service name="CPU"/>
		<service name="IO_MEM"/>
		<service name="IO_PORT"/>
		<service name="IRQ"/>
		<service name="LOG"/>
		<service name="PD"/>
		<service name="RAM"/>
		<service name="RM"/>
		<service name="ROM"/>
	</parent-provides>

	<default-route>
		<any-service> <parent/> <any-child/> </any-service>
	</default-route>

	<start name="linux_nic_drv" ld="no" caps="100">
		<resource name="RAM" quantum="4M"/>
		<config tap="tap0" mac="02:00:00:00:00:00"/>
	</start>

	<start name="timer" caps="100">
		<resource name="RAM" quantum="1M"/>
		<provides> <service name="Timer"/> </provides>
	</start>

	<start name="nic_router" caps="200">
		<resource name="RAM" quantum="10M"/>
		<provides>
			<service name="Nic"/>
			<service name="Uplink"/>
		</provides>
		<config verbose_domain_state="yes" verbose_packets="no">
			<domain name="uplink" interface="10.0.2.2/24" gateway="10.0.2.1">
[...]

PR #4343, opened 2021-12-13
fix: use correct default value for nameserver

sergey-platonov:
Currently, the default path for the `nameserver` file is fixed and it's `/socket/nameserver`. So, if the socket directory is not `/socket` DNS will not work (unless the user will manually set `nameserver` attribute). These changes fix this, now the default path for the `nameserver` takes into account the `socket` path. 

I have tested these changes with the `fetchurl` runscript on `base-nova`. The second commit has changes that demonstrate that.

Issue #4318

### DIFF
--- a/repos/libports/run/fetchurl.inc
+++ b/repos/libports/run/fetchurl.inc
@@ -137,11 +137,11 @@ install_config {
 				<dir name="dev">
 					<log/> <null/> <inline name="rtc">2000-01-01 00:00</inline>
 					<inline name="random">01234567890123456789</inline>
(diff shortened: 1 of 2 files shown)


chelmuth:
I merged the fix as 72bba6bcffa8bf8d0670a56bc475851e7497ee96 with slightly adjusted commit message. You may correct me, but I could not identify the `nameserver` attribute you mentioned in your message.

sergey-platonov:
@chelmuth thank you! Of course, it was supposed to be the `nameserver_file`.

Issue #23, opened 2011-12-20
Enable -Wconversion compiler flag (was mantis id 134)


nfeske:
GCC manpage about the Wconversion option:

Warn for implicit conversions that may alter a value. This includes conversions between real and integer, like `abs (x)' when`x' is `double'; conversions between signed and unsigned, like`unsigned ui = -1'; and conversions to smaller types, like `sqrtf (M_PI)'. Do not warn for explicit casts like`abs ((int) x)' and `ui = (unsigned) -1', or if the value is not changed by the conversion like in`abs (2.0)'. Warnings about conversions between signed and unsigned integers can be disabled by using `-Wno-sign-conversion'.

m-stein:
@nfeske: I would start adapting the base repos according to this flag if you think it's still a good idea!?

m-stein:
Of course, as less prioritized work while I'm not able to proceed with other stuff.

m-stein:
My plan would be to push little chunks of adaption (for instance one commit for util/string.cc) while adding the flag only in my tools.conf until I reach the point when I can add the flag to the default opts without doing any harm. This way, my work would be continuously merged without a lot of re-basing effort.

nfeske:
Yes, that is the perfect approach, @m-stein. Thanks for taking on this issue!

Let me share the experience of my recent work on `-Wsuggest-override`, which was a very similar undertaking.

1. Just like you suggested, I enabled the warning (in base/mk/global.mk) on my local topic branch.

2. I concentrated on one architecture (x86_64) on the generic code first. In the build.conf of my x86_64 build directory, I enabled only the base and os repositories.

3. I repeatedly built the whole tree by issuing `make` without arguments, fixed a warning, and built again. To speed up the repetitive process, the `make again` feature of the build system becomes handy. It omits the dependency-generation stage and executed the build step immediately. I usually pipe the build output into less to make it easier searchable (`make 2>&1 | less`). So you can see the next error quickly w/o scrolling. I found the use of parallel builds more of a hindrance than the benefit because it does - of course - not always show you the result of the compilation unit one just attempted to fix. When using -j4 one sees four errors at once (which is nice) but one has to keep them and their fixes in short-term memory all at once. Since this is quite taxing, it's probably best to disable parallel builds.

4. After the generic parts of the os and base built completely, I focused on the various kernels. For each of the 8 kernels, I executed `make run/log`. This way, core and the base framework is built for each kernel.

[...]

PR #4347, opened 2021-12-14
Refactor VirtIO::Queue implementation a bit.

tworaz:
A while ago I've decided that the current VirtIO::Queue implementation could use a few improvements and cleanups. My original plan was to submit this together with VirtIO block driver, but I currently have no idea when this project will be finished. After some discussions with @nfeske Ive decided to submit the VirtIO::Queue refactoring patch on its own. It does not introduce any major changes in the public API of the class. The patch mainly tries to improve internal implementation of the queue. IMO the resulting code should be easier to read and understand.

The only notable API change this patch introduces is removal of request_irq argument from most public VirtIO::Queue functions. None of the drivers submitted thus far use it and this is unlikely to change in the near future. Most drivers do not even specify this argument when using VirtIO::Queue functions and just use the default value. As such most drivers don't even need to be modified.

Besides easier to read code this refactored version should also be a bit easier to test. Previous version had two code paths which implemented descriptor chaining. The path used by write_data_read_reply function had a bug in it which @nfeske discovered while working on #23. The fix for it was trivial, but testing it, not so much.

The refactored version of VirtIO::Queue should use a shared descriptor writing code path. As result testing descriptor chaining can be easily done using existing virtio_nic driver. For example, apply the following patch to the Genode tree:
```
--- a/repos/os/recipes/raw/drivers_nic-virt_qemu/drivers.config
+++ b/repos/os/recipes/raw/drivers_nic-virt_qemu/drivers.config
@@ -37,7 +37,7 @@

        <start name="virtio_mmio_nic">
                <resource name="RAM" quantum="5M"/>
-               <config mode="uplink_client"/>
+               <config mode="uplink_client" verbose="true" tx_buffer_size="16" tx_queue_size="1024"/>
                <route>
                        <service name="Platform">
                                <child name="platform_drv"/>
```
This should configure the virtio_nic driver to use 1024 16 byte buffers for transmitting packets. 16 bytes is not enough to transmit single packet produced by any of the existing Genode network related run scenarios (ex. run/ping). The virtio_nic driver should still work however since in such case it'll just use multiple descriptors (descriptor chaining) to transfer each TCP/IP packet.


nfeske:
Thanks a lot for this quite unexpected follow-up!

The `Queue` change is very much appreciated. The code becomes much more readable indeed.

It also great to see all those pesky `Genode::` prefixes go. :-)

I merged both commits to staging.

PR #4354, opened 2021-12-18
Squash a few extra Wconversion related warnings

tworaz:
This is a tiny collection of fixes related to recently enabled Wconversion warning.

As far as I can tell Wimplicit-int-conversion warnings are unique to clang. I haven't found a way to produce similar warnings in GCC. I tried to describe those as best as I could, so hopefully it should be clear why the fixes are needed.

The patch related to Wsign-conversion does not fix all the warnings enabled by this flag. All of the warnings fixed and quite a few additional ones can be also seen with GCC by passing -Wsign-conversion flag.


nfeske:
Thank you for the welcome fixes. I merged your commit series to staging.

nfeske:
The fixes entered the master branch in the meanwhile.

PR #4358, opened 2021-12-21
add golang as target for compilation

tor-m6:
introduce makefile target for .go files.

Probably for go packages (compiled to .o files) we need to have some reference infrastructure - they should be provided for gccgo compilation via -I flag (just like C provides .h files).


nfeske:
Thanks for the pull request. I merged your commit as d109015160d0171fc724c29d7be235990342664e (with a minor whitespace adjustment) to staging.

nfeske:
Fixup https://github.com/genodelabs/genode/commit/e6231794f36a481afc9edb058df342f0c8ba1f6c tweaks the commit a bit to make it more consistent with the other languages.

@tor-m6 Could you please give this version a try to confirm that the fixup does not break anything?

tor-m6:
yes, I can confirm that it allow compilation of .go

PR #4370, opened 2022-01-09
fix `socket_fs_socket`: socket type

sergey-platonov:
The small bug that was treating `SOCK_STREAM` sockets as UDP if
additional flags were set.

### DIFF
--- a/repos/libports/src/lib/libc/socket_fs_plugin.cc
+++ b/repos/libports/src/lib/libc/socket_fs_plugin.cc
@@ -1033,9 +1033,9 @@ extern "C" int socket_fs_socket(int domain, int type, int protocol)
 		error(__func__, ": socket fs not mounted");
 		return Errno(EACCES);
 	}
-
-	if (((type&7) != SOCK_STREAM || (protocol != 0 && protocol != IPPROTO_TCP))
-	 && ((type&7) != SOCK_DGRAM  || (protocol != 0 && protocol != IPPROTO_UDP))) {
+	int low_type = type & (SOCK_STREAM | SOCK_DGRAM);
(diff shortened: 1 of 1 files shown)


chelmuth:
Can I assume you have an application setting the socket type in conjunction with any of the SOCK_NONBLOCK and SOCK_CLOEXEC option? While masking all option bits the remainder of the code does not consider the options (esp. not SOCK_NONBLOCK). May this become an issue?

chelmuth:
I merged the slightly adjusted commit cff8567752a312a7ae57b883f337618e311522bb.

sergey-platonov:
Hi @chelmuth 
Yes, that is exactly what I'm doing: porting a new version of `libevent`. Thank you for the heads up!

I have a couple of questions though. 

Is it ok to call `fcntl(fd, F_SETFL, O_NONBLOCK)` after the socket is created? Will that make the socket non-blocking?

`libevent` supports several backends (`epoll`, `poll`, `select`, etc). As far as I understand I can choose between `poll` and `select` for Genode port. Which one should I use, in your opinion?

chelmuth:
> Is it ok to call `fcntl(fd, F_SETFL, O_NONBLOCK)` after the socket is created? Will that make the socket non-blocking?

AFAIK non-blocking TCP sockets work for many cases, e.g., Falkon and gRPC. Currently, the application (or libevent library in your case) has to call `fcntl()` explicitly after socket creation and should not rely on the `SOCK_NONBLOCK` flag. You may add support for this and send a patch for review.

> `libevent` supports several backends (`epoll`, `poll`, `select`, etc). As far as I understand I can choose between `poll` and `select` for Genode port. Which one should I use, in your opinion?

We implement `select()` and have a wrapper to provide `poll()`. That said, I'd expect best results with a select-based backend in libevent.

PR #4422, opened 2022-02-11
Fix handling of empty global static constructors array

tworaz:
This does not affect default Genode builds as far as I can tell. There
is always at least one global static CTOR which seems to be coming from
one of the GCC runtime libs bundled in the toolchain. The problem became
visible for me only after I've replated GCC runtime with LLVM based
one. In such setup I often see binaries that do not have any static ctors.
Such binaries end up crashing Genode ld.lib.so.

Make sure the code does handle empty constructors array.


nfeske:
Merged as https://github.com/genodelabs/genode/commit/800549a2292103457a00e80889024b94571cbc66 with a minor style fix. (putting the control-flow-altering `return` at the beginning of a new line).

PR #4423, opened 2022-02-11
base: Don't produce .gnu.hash tables for Genode.

tworaz:
Genode linker does not support .gnu.hash tables so they will never be
used. Tell the linker not to bother producing them. This should reduce
the size of Genode ELF files a tiny bit without loosing anything
important in the process.


chelmuth:
Thanks for the clarifying commit. I checked my binaries but those do not contain .gnu.hash sections and the manpage of *ld* tells

> The default is "sysv".

So, did you see binaries with .gnu.hash with your tool chain? If yes, is your ld different from our binutils based tool?

tworaz:
Ah, you might be right. Most of the time I'm using clang+lld to build Genode and this is indeed where the difference is. I just checked GCC based build and indeed there is no .gnu.hash in binaries produced by it.

Anyway, sorry for the noise. I'll drop the PR.

chelmuth:
You got me wrong, I like to clarify the intention to not use fancy .gnu.hash tables. I was just wondering how you got those sections. Please push your branch again and repoen the PR.

tworaz:
As you wish. The patch does make it explicit Genode uses only sysv hash-tables. At the same time if this is already the default in the supported toolchain. LLD without any genode specific patches produces both table types.

I could probably fix my Genode clang/lld port to also only produce sysv tables by default, but I probably will just keep this patch around since I prefer to not change llvm/clang defaults too often. Also adding support for .gnu.hash tables in ld.lib.so is on the list of projects that might be fun to do one day :).

Anyway, feel free to do whatever you wish with the patch.

PR #4424, opened 2022-02-11
Allow Genode to boot on Cortex-A55 based SOCs like RK3566.

tworaz:
Support for booting Genode on RK3566 based systems like Quartz64 A board will be provided in a standalone genode-rockchip repository. The repository is still in WIP state and has not been made public, yet.

This patchset contains two small fixes that should allow the port to boot in SMP mode once its public.


nfeske:
Great news learning about your Rockchip plans!

I'd like to leave the review of this pull request to @skalk.

repos/base-hw/src/bootstrap/spec/arm_64/crt0.s
```diff
@@ -19,11 +19,33 @@
...
+	bic x8, x8, #(1 << 0)
+	msr sctlr_el1, x8
+	isb
+1:
+	mrs x8, sctlr_el2
+	bic x8, x8, #(1 << 0)
+	msr sctlr_el2, x8
+	isb
+.endm
+
```
skalk, line 45: @tworaz I highly appreciate your initiative to be more robust in different boot environments. However, I wonder whether we then would have to check for EL3 too. Moreover, isn't it possible to move this code sequence into high-level C++, e.g., at the beginning of the processor-specific enable_mmu functionality? I would expect all code in between to not interfer with being executed with enabled MMU. For instance on x86, we also typically get booted with 1:1 mappings and enabled MMU. But maybe I oversee some problems here?!
It is just that less people speak ARMv8 assembler, and such branching is better readable in C++ than in assembler. So if it is possible to disable the MMU later, I would favor it.

skalk:
@tworaz I very much like your carefully written commit 25912ce and woukd like to merge it instantly. The other commit related to the MMU disablement I've commented inline, and would like to first discuss an alternative approach. Thank you very much for your contributions!

repos/base-hw/src/bootstrap/spec/arm_64/crt0.s
tworaz: I'll have to try if it works. I did originally try to disable MMU at the beginning of enable_mmu, but it did not work. The primary core booted fine, but secondary ones still got stuck somewhere around ttbr write calls. At least it seemed like it since I don't have any good debugger that would allow me to pinpoint the exact location where things break.

The thing is this happened before I discovered the problem addressed by the second patch. It might be that disabling MMU at the beginning of enable_mmu call will work fine now. Without the MPIDR patch all the cores where considered as core 0 and used the same stack from crt0.s.

I should have some time in the evening to run some tests.

tworaz:
@skalk Great. As I wrote in the other comment I'll see if I can experiment once more with disabling MMU in C++ code at the beginning of enable_mmu function. If it works, I'll update the patch.

repos/base-hw/src/bootstrap/spec/arm_64/crt0.s
tworaz: @skalk So I run some tests yesterday and things are not so simple. Moving this code into C++ enable_mmu function did not work. At least not for SMP use case. With mmu disabled at the beginning of enable_mmu function things boot fine with NR_OF_CPUS=1. With NR_OF_CPUS>1 all the secondary cores get stuck somewhere before enable_mmu. They do seem to get into bootstrap's init function, but not further than that.

So next I've tried experimenting with the _mmu_disable call this patch adds in crt0.s. What I've managed to establish is the mmu needs to be disabled for primary core before the _crt0_fill_bss_zero section in this file. If the primary core disables MMU after clearing BSS, secondary cores will not start as they should. Secondary cores themselves can disable MMU at the beginning of enable_mmu C++ function and things will continue to work.

Right now I'm not really sure what to make of it. This smells like some sort of problem with dcache. Also it might be that u-boot for quartz64 does not actually setup identity mappings for the MMU. If it did that I'd expect for primary core to also fail to boot.

skalk: @tworaz thank you for investigating the problem again. I see that it is problematic if the boot environment did not consistently setup the different cores. So I agree with disabling the MMU within the assembler path. However, after looking again at the commit, I've seen that you unconditionally disable the MMU of EL2, also if you detected EL1, am I right? I would think this will cause a fault, doesn't it? I'll rework your commit slightly to address EL1-EL3 and only disable the MMU of the actual exception level. Everything else can be addressed at the higher levels, when preparing less privileged levels.

tworaz: @skalk  You're right the patch I've pushed has EL2/EL1 branches swapped. I made this mistake originally when cleaning up the code. Then I found the bug, made a fixup, but apparently this fixup targeted the wrong patch. The patch I cheery picked to this review did not have that fixup since it was squashed into another patch. I've just pushed the correct the patch.

The goal is to disable MMU for both EL1 and EL2 when booting in EL2 and only in EL1 when starting in this exception level. I can do something similar for EL3, but I have no HW to actually test it on. If you want I can only disable MMU for the exception level in which the code boots and handle higher levels in C++ code.

If you have the time feel free to rework this patch to something you feel is better suited for integrating into main Genode branch. I probably won't have the time to work on Genode related stuff in the next 2 days.

skalk:
@tworaz please have a look at skalk/genode/commit/62232b406fd0ee65da7bee042964a7c67a19be4f I've already tested the boot on our ARMv8 platforms, but would like to wait for your feedback, whenever you time permits it to test it on RK3566.

tworaz:
@skalk I did a quick check using Genode smp test and your patch version works. All cores boot fine and the test seems to pass.

skalk:
@tworaz thank you for the quick reaction, I've put it into the queue of commits for staging. As soon as the QA has reviewed it too, it'll be available.

PR #4533, opened 2022-06-15
Allow Genode VirtIO drivers to be used on RISC-V

tworaz:
This is a prerequisite for https://github.com/genodelabs/genode-riscv/pull/3. The fb, input and nic drivers don't need much effort. Adding RISC-V spec files is enough to make them build and run.

The virtdev_rom driver needs a tiny bit more work since some of the constants it hard codes are different on RISC-V. Still, its not that hard to also make it work.


ssumpf:
@chelmuth: I have tested the VirtIO drivers, works like a charm, and pushed to riscv staging. Since 19138e7 and 1b301b4 are required by https://github.com/genodelabs/genode-riscv/pull/3 would you mind giving it a try over the weekend?

chelmuth:
I merged the commits to staging.

chelmuth:
Merged to master.

PR #4560, opened 2022-07-14
rom_filter: add <node> for sub_nodes outside of <inline>

Cochimetl:
When using rom_filters to create ROMs with more than one layer of sub_nodes, this is (as far as I can see) only possible in <inline> blocks. However, you can't add <if><has_value/><then> nodes in <inline> blocks, since they are copied inline instead of evaluated (which is correct).

This is annoying in situations like this:
```xml
<if><has_value input="A" value="B"/> <then>
  <inline>
    <X>
      <Y>
        <Z attr="yes"/>
      </Y>
    </X>
  </inline>
</then><else>
  <inline>
    <X>
      <Y>
        <Z attr="no"/>
      </Y>
    </X>
  </inline>
</else></if>
```

WIth this commit, we could instead write:
```xml
<node type="X">
  <node type="Y">
    <if><has_value input="A" value="B"/> <then>
      <inline>
        <Z attr="yes"/>
      </inline>
    </then> <else>
      <inline>
        <Z attr="no"/>
      </inline>
    </else> </if>
  </node>
</node>
```

This is especially useful for things like configs of sub_inits, where you have to write the enitre start node of a component twice if you wish to use if statements for things like its routes or config.


Cochimetl:
I've been thinking about this, and I believe it can be done more elegantly. It should be possible to simply treat all nodes that aren't of an accepted type of rom_filter (if, inline, attribute, attribute and input) as subnodes. That way, a new accepted type isn't necessary. 
Also, we could iterate over the attributes of such nodes and inset them directly.
As I see it, that would make attribute and inline nodes unneeded except for the top node for attribute nodes, and for writing rom_filter accepted nodes that aren't supposed to be used by rom_filter itself.

I'll come back with an updated commit later.

nfeske:
@Cochimetl I like your suggestion. Thank you for bringing it up.

Your commit looks good. The only thing that's missing would be a short mentioning of the new feature in the accompanied README (https://github.com/genodelabs/genode/blob/master/repos/os/src/server/rom_filter/README). Would you like to add a short paragraph there?

In order to merge your patch, we need your consent that you are fine with Genode Labs offering your contribution as part of the its commerical license offering of the Genode OS Framework. You can grant us your permission by sending a signed copy of the Genode contributor's agreement (https://genode.org/community/contributions#Genode_Contributors_Agreement) to licensing@genode-labs.com.

nfeske:
>  It should be possible to simply treat all nodes that aren't of an accepted type of rom_filter (if, inline, attribute, attribute and input) as subnodes. That way, a new accepted type isn't necessary.

I would prefer to stay conservative and not mix the node types interpreted by the ROM filter with the processed node types. Otherwise, any future extension of the ROM filter may break existing use cases.

Cochimetl:
> I would prefer to stay conservative and not mix the node types interpreted by the ROM filter with the processed node types. Otherwise, any future extension of the ROM filter may break existing use cases.

I see your point. I'll leave it be and just add a README paragraph

Cochimetl:
README is changed.

chelmuth:
I'm wondering if your change supports attributes in the nodes like follows.

```xml
<node type="X" attrib="valueX">
  <if><has_value input="A" value="B"/> <then>
    <inline>
      <Z attr="yes"/>
    </inline>
  </then> <else>
    <inline>
      <Z attr="no"/>
    </inline>
  </else> </if>
</node>
```
results in (first branch)
```xml
<X attrib="valueX">
  <Z attr="yes"/>
</X>
```

nfeske:
> I'm wondering if your change supports attributes in the nodes like follows.

That is already covered by the existing `<attribute>` node.

nfeske:
I just added the commit with minor tweaks as https://github.com/genodelabs/genode/commit/3bcf0e6059939ecc557b88331dc1aced64ba4baa to staging.

- Abbreviated the title line of the commit to 50 characters (recommended good practice)
- Changed spelling of "sub_node" to "sub node" in the readme and commit message
- Added reference to the corresponding issue number to the commit message

@Cochimetl Do you approve?

Cochimetl:
> @Cochimetl Do you approve?
Yes, this is fine by me.

PR #4573, opened 2022-07-25
audio_in_session: fix wrong buffer in content

Cochimetl:
When trying to fill the unspecified part of the packet content with zeroes, the memory behind the given data buffer is filled instead. This will lead to page faults in many circumstances.


chelmuth:
@cnuke @cproc are you aware of this PR? What's your opinion?

cnuke:
@chelmuth yes and it is a solid fix.

Cochimetl:
Thanks, I'll close the pull request now.

PR #4597, opened 2022-08-31
Usb ctrl timeout jiffies

Cochimetl:
Sometimes urb timeouts on smart card readers (which were supposed to be 1 second) were triggered almost immediately. Sometimes, they were not triggered at all.
Forcing a jiffies update before using them (as described in a comment in clocksource.c) seems to solve this problem.

### DIFF
--- a/repos/dde_linux/src/lib/lx_emul/usb.c
+++ b/repos/dde_linux/src/lib/lx_emul/usb.c
@@ -23,6 +23,7 @@
 
 #include <lx_emul/shared_dma_buffer.h>
 #include <lx_emul/task.h>
+#include <lx_emul/time.h>
 #include <lx_user/init.h>
 #include <lx_user/io.h>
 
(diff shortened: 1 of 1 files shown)


chelmuth:
Please help me because I could not find the `lx_emul_force_jiffies_update()` function anywhere.

alex-ab:
The function got removed in #4540 as the only user (intel_fb) did not needed it anymore.

Cochimetl:
> The function got removed in #4540 as the only user (intel_fb) did not needed it anymore.

Yes, it's that one.

chelmuth:
Thanks for the clarification. So, it is unclear if the change in this PR is still required in the current implementation. And further, if it is required: *How could a jiffies update be enforced with NO_HZ?*

chelmuth:
> So, it is unclear if the change in this PR is still required in the current implementation. And further, if it is required: _How could a jiffies update be enforced with NO_HZ?_

@alex-ab @Cochimetl any ideas?

alex-ab:
> > So, it is unclear if the change in this PR is still required in the current implementation. And further, if it is required: _How could a jiffies update be enforced with NO_HZ?_
> 
> @alex-ab @Cochimetl any ideas?

As you said, a re-evaluation is required, whether the issue still persists on current master/staging. If it becomes clear, another form of solution is required. The former lx_emul_force_jiffies_update() implementation may also be not applicable anymore, since it was based upon linux tracing/code evaluation in the periodic mode instead of the now used NO_HZ_IDLE mode of all ported drivers.

chelmuth:
Thanks @alex-ab. I close this PR as it's unknown if the issue persists with Genode 22.08. @Cochimetl please reopen with steps to reproduce if the issue reappears.

PR #4711, opened 2022-12-28
Bring in vfs_example to main repo

sid-agrawal:
I found the vfs_examples[1] very useful, but I couldn't find them in the main repo. In this pull-request, I am bringing them into the main repo. If there are already there, can you point me to them, and we can drop this PR :)

Note:
* I have to make changes to the cbe port to make this work. I am not sure if there was a cleaner way to make this work.dd

[1] http://genodians.org/m-stein/2021-06-21-vfs-1
[2] https://github.com/genodelabs/genode


nfeske:
The VFS examples were not meant to be included in the main repository in the first place. They exist merely as supplemental material of @m-stein's postings at Genodians.org.

If you feel that the examples are in need further curation and convenience (like your vscode config), you may consider creating an independent repository and documentation.

For Genode's main repository, we are actually following the opposite direction - removing parts that are non-essential for the base system - e.g., by moving code over to genode-world. This way, the main repository retains a manageable scope.

sid-agrawal:
Got it. Thanks!

Issue #4756, opened 2023-02-08
Broken USB device name


rolandbaer:
In the usb_block_drv the log message "Found USB device:" shows a broken manufacturer and product name like "ds2o (ds2oCASCI)" instead of "disk2go (disk2go CLASSIC II)".

It looks like the encoding is wrong.

chelmuth:
As suspected in #4757, the issue is caused by the use of the `usb_string()` Linux kernel function in *dde_linux/src/lib/lx_emul/usb.c*. From my perspective it would be better to fix it there and return an unmodified string descriptor to a STRING request (maybe retrieved by `usb_get_descriptor()`).

@ssumpf what's your opinion?

ssumpf:
@chelmuth: We wrote the `usb_block_drv` ourselves, it's not dde_linux. All strings given by USB hardware should be UTF-16 (as defined by the standard). Most of the times it should be sufficient to output the lower byte as UTF-8 and skip the upper one. One how this is implemented in `usb_block_drv` we would have to ask @cnuke.

chelmuth:
@ssumpf Don't you agree *repos/dde_linux/src/lib/lx_emul/usb.c* used by *pc_usb_host_drv* should return a string descriptor with UTF-16 strings on a `Usb::Packet_descriptor::STRING` request (see *repos/os/include/usb_session/usb_session.h*)?

cnuke:
The `usb_block_drv` driver uses the implementation from `Usb::String::to_char` from *usb/types.h*. Since this commit alteres the `string` type to *utf-8* that appears to solve the problem.

ssumpf:
@chelmuth: Okay, the `usb_string` function converts the descriptor to UTF-8 (https://elixir.bootlin.com/linux/v5.14.21/source/drivers/usb/core/message.c#L967) `Usb::String` more or less a nop. One could use the `usb_get_descriptor` function in order to retrieve a raw string descriptor (even though the comment says otherwise). Is there a reason we require UTF-16 for strings? Things like VirtualBox6 well get raw descriptors through control messages, and therefore, receive UTF-16 strings. Who else needs it?

chelmuth:
[...]

PR #4772, opened 2023-02-24
usb string utf-16

rolandbaer:
In the usb_block_drv the log message "Found USB device:" showed a broken manufacturer and product name.

This is a second version to fix the issue, this time by handling the usb string as utf16.

fixes https://github.com/genodelabs/genode/issues/4756


chelmuth:
I merged the fix to staging.

PR #4773, opened 2023-02-24
Documentation fixes

rolandbaer:
fixed some mistakes I found in the documentation of file terminal and vfs block.

### DIFF
--- a/repos/gems/src/server/file_terminal/README
+++ b/repos/gems/src/server/file_terminal/README
@@ -2,14 +2,12 @@ File terminal is a service that provides a terminal-session interface and
 redirects the terminal input and output to a file.
 
 ! <config>
-!   <policy label_prefix="client1" filename="test.txt" />
-!   <policy label_prefix="client2" filename="file.dat" io_buffer_size="4K"/>h
-!   <libc stdout="/dev/log">
-!     <vfs>
-!       <dir name="dev"> <log/> </dir>
-!       <fs/>
-!     </vfs>
-!   </libc>
+!   <policy label_prefix="client1" filename="test.txt"/>
+!   <policy label_prefix="client2" filename="file.dat" io_buffer_size="4K"/>
+!   <vfs>
+!     <dir name="dev"> <log/> </dir>
+!     <fs/>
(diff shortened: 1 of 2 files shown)


repos/gems/src/server/file_terminal/README
```diff
@@ -2,14 +2,12 @@ File terminal is a service that provides a terminal-session interface and
 redirects the terminal input and output to a file.
 
 ! <config>
-!   <policy label_prefix="client1" filename="test.txt" />
-!   <policy label_prefix="client2" filename="file.dat" io_buffer_size="4K"/>h
-!   <libc stdout="/dev/log">
```
chelmuth, line 7: This removes the libc stdout configuration. I guess this is not intended.

rolandbaer: This is a snippet of the configuration I used to write log messages to a file:
  <start name="file_terminal">
    <binary name="file_terminal"/>
    <resource name="RAM" quantum="3M"/>
    <provides> <service name="Terminal"/> </provides>
    <config>
      <policy label_prefix="terminal_log" filename="log.txt" /> 
      <vfs>
        <dir name="dev"> <log/> </dir>
        <fs/>
      </vfs>
    </config>
    <route>
      <service name="File_system"> <child name="usb_stick_fs" /> </service>
      <any-service> <parent/> </any-service>
    </route>
  </start>

When I had had the libc node around the vfs node I got a message that it is deprecated (don't know the exact message by heart and havo no system here to run the system on it). With removing the lilbc node the deprecated message disappeared, so I thought this would be the right solution and the doc is outdated.

rolandbaer: I had a look at the log_core.run sample and added the libc node again, but as a own line.

By the way, this was the warning that leaded me to the solution above:
[init -> file_terminal] Warning: '<config> <libc> <vfs/>' is deprecated, please move to '<config> <vfs/>'

chelmuth:
Squashed and merged as 84d17eea96fad57b4baaef8633f0a32519e8397d. Thank.

PR #4816, opened 2023-04-19
base: add Xml_node::for_each_attribute

Time0o:
Adds a function that allows iterating over an XML node's attributes without relying on `Xml_node::Nonexistent_attribute`.


Time0o:
Closing in favor of https://github.com/genodelabs/genode/issues/4817

PR #4919, opened 2023-06-09
depot_download_manager: fix verifying failed download

Time0o:
Dear Genodians,

I have noticed a small issue where a failed `fetchurl` request would result in attempted package verification instead of correctly signalling a failed download. This does not seem to happen on all platforms. I believe this is a bug and this should fix it.


nfeske:
Thank you for the report and the much appreciated fix. I just merged your patch to staging.

PR #4967, opened 2023-07-14
remove duplicates for libs to visit during make

tor-m6:
I found a strange problem (use make 4.1). After some long work (which does not involve genode build system but compile code in build dir) it start to generate wrong libdeps file: some generated variables like DEP_A_base-nova-common refer to itself, directly or indirectly. Seems that this is because of LIBS variable somehow start to be modified from "bottom" recursive make invocation (still dont understand what is a trigger: I try to remove libdeps/progress.log from build, check shell environment variables - but this does not helps).
This is not a first time - every last years advances to new version works for some time for me, and then hit this unfixable problem forcing me to make another clean installation.

I make small patch which remove dup LIBS entries as well as LIB which should be build in invocation.
Seems that this could improve build performance (do not call unneeded recursive make invocations which goto already_visited target)

### DIFF
--- a/repos/base/mk/dep_lib.mk
+++ b/repos/base/mk/dep_lib.mk
@@ -135,14 +135,12 @@ ifneq ($(DEP_MISSING_PORTS),)
 	@(echo "MISSING_PORTS += $(DEP_MISSING_PORTS)"; \
 	  echo "") >> $(LIB_DEP_FILE)
 endif
-	@for i in $(LIBS_TO_VISIT); do \
-	  $(MAKE) $(VERBOSE_DIR) -f $(BASE_DIR)/mk/dep_lib.mk REP_DIR=$(REP_DIR) LIB=$$i; done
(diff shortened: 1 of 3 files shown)


nfeske:
Sorry for attending this pull request so late.

Thanks sharing your patch. Admittedly, I struggle to relate to the problem it is supposed to solve, which is the reason I felt unable to attend it. Ultimately, I would rather not add any complexity to the build system without a definite (and reproducible) reason.

Should similar problems as you reported occur in the future, I hope that the information you provided may be helpful to people searching the issue tracker.

Would you be fine with closing this issue?

tor-m6:
this is up to you
while I still think that there are no reasons to duplicate makefile visits to some directories, this slowdown overall compilation at least.

nfeske:
Thank you for the response. Should we decide to optimize this aspect of the build system sometime in the future, we might open a dedicated issue. I'm closing this one.

PR #4993, opened 2023-09-07
fb_sdl: update to use sdl2

mewmew:
Following the official migration guide of SDL [[1]], the fb_sdl framebuffer driver was update from SDL1 to SDL2.

The sdl2 port in world/src/lib/sdl2 is used.

Since SDL1 is in maintenance mode [[2]], support for other display servers than X11 will never be implemented. In particular, support for Wayland is missing from SDL1.

Fortunately, a port of sdl2 is maintained in genode-world.

As SDL2 is actively developed, it will provide support for modern hardware architectures, and has mature support for Wayland [[3]].

[1]: https://wiki.libsdl.org/SDL2/MigrationGuide
[2]: https://wiki.debian.org/Wayland#SDL1_.28unsupported.29
[3]: https://wiki.debian.org/Wayland#SDL2_.28supported_since_2.0.2.2B-.29

### DIFF
--- a/repos/os/src/drivers/framebuffer/sdl/convert_keycode.h
+++ b/repos/os/src/drivers/framebuffer/sdl/convert_keycode.h
@@ -114,12 +114,12 @@ static inline Input::Keycode convert_keycode(int sdl_keycode)
 	case SDLK_F15:          return KEY_F15;
 
 	/* special keys */
-	case SDLK_PRINT:        return KEY_PRINT;
-	case SDLK_SCROLLOCK:    return KEY_SCROLLLOCK;
+	case SDLK_PRINTSCREEN:  return KEY_PRINT;
+	case SDLK_SCROLLLOCK:   return KEY_SCROLLLOCK;
 	case SDLK_MENU:         return KEY_MENU;
 
 	/* key state modifier keys */
-	case SDLK_NUMLOCK:      return KEY_NUMLOCK;
+	case SDLK_NUMLOCKCLEAR: return KEY_NUMLOCK;
 	case SDLK_CAPSLOCK:     return KEY_CAPSLOCK;
 	case SDLK_RSHIFT:       return KEY_RIGHTSHIFT;
 	case SDLK_LSHIFT:       return KEY_LEFTSHIFT;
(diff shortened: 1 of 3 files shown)


mewmew:
Screenshot using `goa run` with the `fb_sdl` framebuffer driver on Wayland.

fb_sdl_wayland

chelmuth:
I tested your commit with a quite dated installation (libsdl2 version 2.0.8+dfsg1-1ubuntu1.18.04.4) and it worked to some extent. Unfortunately, the windows resize is no longer propagated as screen/framebuffer-resolution change but just scales the windows contents. Does this feature still work for you?

mewmew:
> I tested your commit with a quite dated installation (libsdl2 version 2.0.8+dfsg1-1ubuntu1.18.04.4) and it worked to some extent. Unfortunately, the windows resize is no longer propagated as screen/framebuffer-resolution change but just scales the windows contents. Does this feature still work for you?

Wow, really good thing you tested this better than I did @chelmuth. It turns out that the window resize events were never handled at all (see https://github.com/genodelabs/genode/pull/4993/commits/e6335a16b5e79c32a8793bae388b37526fc9e8e1 for a fix to the embarrassing typo).

In essence, `event.type` was compared against `SDL_WINDOWEVENT_RESIZED` (which is incorrect). The correct check would be to compare `event.window.event` against `SDL_WINDOWEVENT_RESIZED` (`event.type` is `SDL_WINDOWEVENT`).

Having fixed said typo, the resize code turned out to not work as expected in SDL2. The reason for this is that SDL2 has added support for multiple windows. So, when `_sdl_screen.construct` was invoked upon window resize events, a *new* window would be created each time (with the new window dimensions), instead of resizing the old window.

This issue has been addressed in 49d35cbe6dd3364d3132148178fdd26af0e6b045, which refactors the code to only construct a single window, and then reconstruct the `SDL_Surface` and `SDL_Texture` of `_sdl_screen` upon window resize events.

Please try again with these two commits added, and see if the behaviour is fixed on your end too.

Cheerful regards,
Robin

chelmuth:
Great, now it works like a charm :+1: I just added fixup b1a6cdb349e302e88b62f49b8401b0fe474170b3 to prevent build errors on Ubuntu 20.04.

chelmuth:
> SDL2 has added support for multiple windows. So, when `_sdl_screen.construct` was invoked upon window resize events, a _new_ window would be created each time (with the new window dimensions), instead of resizing the old window.

That's quite interesting and could become useful when prototyping advanced multi-monitor support in Nitpicker and wm...

mewmew:
> Great, now it works like a charm 👍 I just added fixup https://github.com/genodelabs/genode/commit/b1a6cdb349e302e88b62f49b8401b0fe474170b3 to prevent build errors on Ubuntu 20.04.

Great! Closing this PR as the commits have been merged to staging.

Cheers,
Robin

PR #5784, opened 2026-01-07
fb_sdl: SDL2/3 compatability conversion

b0inbot:
Converts the SDL2 fb_sdl codebase to SDL3 via an sdl2/sdl3.h shim.

Confirmed working on stock SDL2 Ubuntu LTS 24.04. Confirmed working
on manually compiled SDL3 on Ubuntu LTS 24.04. make run/demo
BOARD=linux KERNEL=linux runs and is visually correct.

### DIFF
--- a/repos/os/src/driver/framebuffer/sdl/target.mk
+++ b/repos/os/src/driver/framebuffer/sdl/target.mk
@@ -2,8 +2,17 @@ TARGET   = fb_sdl
 LIBS     = lx_hybrid blit
 REQUIRES = linux
 SRC_CC   = main.cc
-LX_LIBS  = sdl2
-INC_DIR += $(PRG_DIR) /usr/include/SDL2
+SDL_INC  = /usr/include/SDL2
+SDL_LIB  = sdl2
+
+ifeq ($(GENODE_FB_SDL_VERSION),3)
+CC_OPT += -DSDL3
+SDL_INC = /usr/include/SDL3
+SDL_LIB = sdl3
+endif
+
+LX_LIBS  = $(SDL_LIB)
+INC_DIR += $(PRG_DIR) $(SDL_INC)
 
 # accept conversion warnings caused by libgcc's include/arm_neon.h
 CC_OPT  += -Wno-error=narrowing \
--- a/repos/os/src/driver/framebuffer/sdl/convert_keycode.h
+++ b/repos/os/src/driver/framebuffer/sdl/convert_keycode.h
@@ -33,7 +33,11 @@ static inline Input::Keycode convert_keycode(int sdl_keycode)
 	case SDLK_PAUSE:        return KEY_PAUSE;
 	case SDLK_ESCAPE:       return KEY_ESC;
 	case SDLK_SPACE:        return KEY_SPACE;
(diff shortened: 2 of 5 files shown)


nfeske:
Splendid! That's exactly what I had in mind. :+1: 

I just gave your commit a quick try. It works fine with SDL2. Really cool!

To ease the switch between both versions, the fixup https://github.com/nfeske/genode/commit/fb9ff4d2dc5fc8485bb4d9700cb08f96adfa7086 allows for controlling the SDL version via the SPECS build variable. So you can enable the SDL3 build without any patch by adding the following line to your build directory's etc/specs.conf file:
```
SPECS += sdl3
```
Note that the SDL library is not captured as dependency of fb_sdl. So when switching between both versions, one has to manually remove the driver/framebuffer/sdl from the build directory before building.

Please feel free to squash the fixup to your commit.

nfeske:
Nice! BTW, the `ifeq` in the target.mk could now be simplified to `ifeq ($(GENODE_FB_SDL_VERSION),3)`.

Maybe @ssumpf has an idea about your remaining open question about the location of the libpulsecommon shared lib on the host? Citing your commit message:

> Running on my own system requires LD_LIBRARY_PATH=/usr/lib64/pulseaudio/
within the make invocation. Fixable somewhere I'm sure. I tried LIB_DIR
and such within the targets.mk.
>
> ```
> [Genode] init -> drivers -> fb_sdl: error while loading shared libraries: libpulsecommon-17.0.so:
> cannot open shared object file: No such file or directory
> $ find /usr/ -iname libpulsecommon-17.0.so
> /usr/lib/pulseaudio/libpulsecommon-17.0.so
> /usr/lib64/pulseaudio/libpulsecommon-17.0.so
> ```

ssumpf:
> Nice! BTW, the `ifeq` in the target.mk could now be simplified to `ifeq ($(GENODE_FB_SDL_VERSION),3)`.
> 
> Maybe @ssumpf has an idea about your remaining open question about the location of the libpulsecommon shared lib on the host? Citing your commit message:

Maybe some additional environment variables need to be forwarded in `Native_pd_component::_start` (_base-linux/src/core/native_pd_component.cc_)? Currently only `LD_LIBRARY_PATH` and `XDG_RUNTIME_DIR` are forwarded.

repos/os/src/driver/framebuffer/sdl/target.mk
```diff
@@ -2,8 +2,17 @@ TARGET   = fb_sdl
 LIBS     = lx_hybrid blit
 REQUIRES = linux
 SRC_CC   = main.cc
-LX_LIBS  = sdl2
-INC_DIR += $(PRG_DIR) /usr/include/SDL2
+SDL_INC  = /usr/include/SDL2
+SDL_LIB  = sdl2
+
+ifeq ($(GENODE_FB_SDL_VERSION),3).
```
nfeske: The trailing dot is wrong.

nfeske:
That looks excellent! I merged your commit to our staging branch now. Thank you very much for this contribution, and also for the pleasant communication. :)

Issue #5088, opened 2024-01-08
ahci_drv: devices not reported


cproc:
When I tried to run Sculpt on a PC with an ASUS Q170T main board, the AHCI driver did not report the connected SATA SSDs, but showed them as `off` in the log:

```
[init -> drivers -> dynamic -> ahci_drv] --- Starting AHCI driver ---
[init -> drivers -> dynamic -> ahci_drv] version: major=0x1 minor=0x301
[init -> drivers -> dynamic -> ahci_drv] command slots: 32
[init -> drivers -> dynamic -> ahci_drv] native command queuing: yes
[init -> drivers -> dynamic -> ahci_drv] 64-bit support: yes
[init -> drivers -> dynamic -> ahci_drv] number of ports: 6
[init -> drivers -> dynamic -> ahci_drv] 		#0: off (ATA)
[init -> drivers -> dynamic -> ahci_drv] 		#1:  off (unknown device signature)
[init -> drivers -> dynamic -> ahci_drv] 		#2:  off (unknown device signature)
[init -> drivers -> dynamic -> ahci_drv] 		#3: off (ATA)
[init -> drivers -> dynamic -> ahci_drv] 		#4:  off (unknown device signature)
[init -> drivers -> dynamic -> ahci_drv] 		#5:  off (unknown device signature)
```

The error symptom in the driver is that setting the `Cmd::Fre` bit to 0 does not cause the `Cmd::Fr` bit to change to 0 as well:

https://github.com/genodelabs/genode/blob/c4679e7af644f2ff8d07b9b48bc2c79eb57a79cc/repos/os/src/drivers/ahci/ahci.h#L880-L885

The problem appears to be related to the port reset at the beginning of the `Port()` constructor:

https://github.com/genodelabs/genode/blob/c4679e7af644f2ff8d07b9b48bc2c79eb57a79cc/repos/os/src/drivers/ahci/ahci.h#L549-L555

When I tried the `Cmd::Fre` bit change before this `reset()` call, the `Cmd::Fr` bit changed to 0 as expected.

The AHCI specification mentions that "software should write all 1s to the PxSERR register to clear any bits that were set as part of the port reset.", which is done after the second port reset in the `init()` function, but not after the first port reset in the `Port()` constructor:

https://github.com/genodelabs/genode/blob/c4679e7af644f2ff8d07b9b48bc2c79eb57a79cc/repos/os/src/drivers/ahci/ahci.h#L856-L860

[...]

PR #5853, opened 2026-04-13
AHCI: Resolve port initialization conflict on boot vs S4 resume

G0retZ:
This PR is vibe-coded with thorough supervision and sanity check.
Though I'm not an expert in C++ nor in drivers development.

Here is the summary of the reasoning behind the changes produced by AI:

> On systems with an IOMMU, SATA ports were logged as "off (ATA)" at boot
> despite the BIOS having successfully initialized the devices.
> 
> Root cause
> ----------
> Port::reinit() unconditionally called reset() (COMRESET via PxSCTL.Det)
> before enable() and before setup_memory() had mapped a Genode-owned FIS
> receive buffer into PxFB.
> 
> At that point PxFB still held the BIOS-allocated physical address and FRE
> (FIS Receive Enable) was left enabled by the BIOS. The device responded to
> COMRESET by posting a D2H Register FIS. The HBA attempted a DMA write to
> the BIOS address, which the IOMMU blocked. As a result FR (FIS Receive
> Running) never cleared. The subsequent setup_memory() → wait_for(FR=0)
> timed out with Polling_timeout, which propagated out of _scan_ports and
> left the port disabled.
> 
> Fix
> ---
> Move the COMRESET out of the unconditional preamble and instead call
> enable() first:
> 
> - On boot and S3 resume the SATA link is already established by the BIOS /
>   powered controller. enable() succeeds immediately and no COMRESET is
>   needed before init().
> 
> - On S4 resume (full power loss) all port registers including PxSSTS reset
>   to zero, so enable() returns false. In that case reset() is issued to
>   re-establish the link. After S4, Cmd::St and FRE are guaranteed 0 by
>   power loss, so the COMRESET cannot trigger an IOMMU-blocked DMA, and
>   setup_memory() inside the subsequent init() will find FR=0 immediately.
> 
> The is_resume parameter (defaulting to false) distinguishes the two
> failure meanings of enable() returning false: an empty port at boot (throw
> immediately, DMA registers never touched) versus a link-down condition at
> S4 resume (reset and retry).
> 
> The resume call site in main.cc is updated to pass is_resume=true.

Tested on my system building Sculpt from the current `master` branch and now I have access to all of my drives.

### DIFF
--- a/repos/os/src/driver/ahci/ahci.h
+++ b/repos/os/src/driver/ahci/ahci.h
@@ -898,13 +898,31 @@ struct Ahci::Port : private Port_base
 		reinit();
 	}
 
-	void reinit()
+	void reinit(bool is_resume = false)
 	{
 		_with_port_mmio([&](Port_mmio &mmio) {
-			reset(mmio);
 
-			if (!enable(mmio))
-				throw 1;
+			if (!enable(mmio)) {
+				/*
+				 * No SATA link. On boot or S3 resume the link is already
+				 * established by the BIOS / powered controller, so a missing
+				 * link means the port is empty.
+				 */
+				if (!is_resume)
+					throw 1;
+
+				/*
+				 * On S4 resume (full power loss) all port registers reset to
+				 * zero including PxSSTS, so the link must be re-established.
+				 * Cmd::St and FRE are guaranteed 0 after power loss, so
+				 * reset() is safe and setup_memory() inside init() will find
+				 * FR=0 immediately.
+				 */
+				reset(mmio);
+
+				if (!enable(mmio))
+					throw 1;
+			}
 
 			stop(mmio);
 
--- a/repos/os/src/driver/ahci/main.cc
+++ b/repos/os/src/driver/ahci/main.cc
@@ -160,7 +160,7 @@ class Ahci::Driver : Noncopyable
 				/* re-start request handling of client sessions */
 				for_each_port([&](auto &port, auto const index, auto) {
 					try {
-						port.reinit();
+						port.reinit(true);
 						port.stop_processing = false;
 						_dispatch.session(index);
 					} catch (...) {


G0retZ:
This addresses the issue #5088

ssumpf:
@G0retZ: [4609a4c](https://github.com/genodelabs/genode/pull/5853/commits/4609a4cbe86fb7c6be2cd41a450f232d5e3c7632) does - as observed by @cproc in #5088 - not call the Port reset during AHCI driver startup. This leaves the question of what will happen in case a BIOS doesn't hand over a half configured controller or in case there is no BIOS (think ARM based SoCs)? I would try to clear `Cmd::St` (unconditionally) and than clear `Cmd::Fre`  (as @cproc suggested) before the Port reset because having FIS receive running during reset is definitely not good. I will send a sketch commit for testing to #5088 and would appreciate if you could test this. Also I would like to continue the discussion in #5088.

nfeske:
Thanks @G0retZ for the continued investigation. Happy to see it getting resolved. Also thanks @ssumpf for responding.

> This PR is vibe-coded with thorough supervision and sanity check.

Thank you for being transparent about it.

As a note of caution, Genode is not open for AI-generated code contributions due to the legal uncertainties of such code. In order to incorporate code into genuine Genode components (those that are marked by the license headers  as being part of Genode) we must ensure that such code does not violate rights of third parties. On this account, the [Genode contributor's agreement](https://genode.org/community/gca.pdf) states "You certify and warrant that your Contributions to Genode Labs’s Products do not violate the intellectual property rights of third parties [...]". This cannot be assumed for code generated by AI models without legal records of their training material. As far as I know, AI service providers do not provide such evidence.

G0retZ:
> Thank you for being transparent about it.
> 
> As a note of caution, Genode is not open for AI-generated code contributions due to the legal uncertainties of such code. In order to incorporate code into genuine Genode components (those that are marked by the license headers as being part of Genode) we must ensure that such code does not violate rights of third parties. On this account, the [Genode contributor's agreement](https://genode.org/community/gca.pdf) states "You certify and warrant that your Contributions to Genode Labs’s Products do not violate the intellectual property rights of third parties [...]". This cannot be assumed for code generated by AI models without legal records of their training material. As far as I know, AI service providers do not provide such evidence.

Ok, I definitely need take this into account next time 👌 
To be more specific this code was generated by Claude Sonet 4.6 as a service under my thorough supervision, so probably it's not eligible for merging as is just because it's generated by a 3rd party service.
Though I'd like to do more contributions in the near future but I'm not experienced enough in C++ and system programming to pull this off easily...

Since I don't have legal expertise here, could you please clarify in more details what counts as an unacceptable "AI generated code" contribution?

What if I will use an AI agent service for advisory and static analysis only, but will type the actual code by my own hands?
Or if I will use [a locally running AI model](https://huggingface.co/bartowski/google_gemma-4-26B-A4B-it-GGUF) that is released under Apache 2.0 or other GPL3 compatible license?

I'd like to know my limits here from Genode perspective 🤔

nfeske:
> Or if I will use [a locally running AI model](https://huggingface.co/bartowski/google_gemma-4-26B-A4B-it-GGUF) that is released under Apache 2.0 or other GPL3 compatible license?

I'm not a lawyer. But I consider learning from any source (book, teacher, experience, internet search engine, AI model) as fine as long as the information source is legal (think of trade secrets).

Code generation is different matter. Legally, regardless of whether running the model locally or as a service, unless you created the model yourself, you cannot know from which data the model parameters had been derived. So one should assume the worst. Common sense. The obvious way to stay safe is to craft code by hand, driven by human intent.

At the end of the day, someone has to assert responsibility, for both technical and legal risks. A maintainer becomes responsible for the code whenever accepting a contribution. Accepting the risks of AI-generated artifacts would be irresponsible for Genode Labs. Hence, we have to keep our code untainted.

G0retZ:
> I'm not a lawyer. But I consider learning from any source (book, teacher, experience, internet search engine, AI model) as fine as long as the information source is legal (think of trade secrets).
> 
> Code generation is different matter. Legally, regardless of whether running the model locally or as a service, unless you created the model yourself, you cannot know from which data the model parameters had been derived. So one should assume the worst. Common sense. The obvious way to stay safe is to craft code by hand, driven by human intent.
> 
> At the end of the day, someone has to assert responsibility, for both technical and legal risks. A maintainer becomes responsible for the code whenever accepting a contribution. Accepting the risks of AI-generated artifacts would be irresponsible for Genode Labs. Hence, we have to keep our code untainted.

Thanks for clarification! 🙏 
I will change my approach to avoid generating code with AI for contributions from now on. I will use AI as a mere learning and reasoning tool, but will craft code by hand 👍

ssumpf:
Fixed in #5088

